Issue #5, opened 2020-05-07
Plotting


willtebbutt:
Stheno [has functionality to make plotting GPs in 1D easy](https://github.com/willtebbutt/Stheno.jl/blob/master/src/util/plotting.jl). It would be great if someone could port this over.

@sharanry this might be a good thing to do to make plotting stuff for the readme / examples easier :)

sharanry:
Thanks for that! Used that to make this.

![anim](https://user-images.githubusercontent.com/19674071/81372748-bb8c5400-9118-11ea-8e38-e8baf030266c.gif)

sharanry:
Slightly better looking with some samples.
![anim](https://user-images.githubusercontent.com/19674071/81373759-1cb52700-911b-11ea-9c38-41880ae88e62.gif)

cpfiffer:
Is there anything cooler looking than a GP plot?

sharanry:
> Is there anything cooler looking than a GP plot?

Hey! Do you have anything in mind?

cpfiffer:
No, sadly. GP plots are the coolest so there can be no improvement.

willtebbutt:
Closing this as I believe it to be resolved. Please re-open if there are improvements still to be made.

nathanaelbosch:
>> Is there anything cooler looking than a GP plot?
>
> No, sadly. GP plots are the coolest so there can be no improvement.

I disagree! What about GP plots with animated samples?
![posterior2](https://user-images.githubusercontent.com/25008422/94360970-933dcb80-00a0-11eb-8fa7-54e41402c4c2.gif)

I made this gif last week, following [Philipp Hennig's Tech Report on "Animating Samples from Gaussian Distributions"](http://mlss.tuebingen.mpg.de/2013/2013/Hennig_2013_Animating_Samples_from_Gaussian_Distributions.pdf). The code to generate the sample paths is fairly generic, but making the actual animation did of course depend on Plots.jl.

I would love to make the code for these animations more accessible. Is this something you would like to include in AbstractGPs.jl? Or what would be the best approach?

devmotion:
Very cool! A colleague of mine created the following webpage based on this idea: http://www.it.uu.se/edu/course/homepage/apml/GP/

[...]

PR #6, opened 2020-05-08
Port plotting utils from Stheno

sharanry:
#5 

Direct port from Stheno to plot 1-D GPs.


Project.toml
```diff
@@ -8,7 +8,9 @@ Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
 FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
 KernelFunctions = "ec8451be-7e33-11e9-00cf-bbf324bd1392"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
```
devmotion, line 11: Depending on RecipesBase should be sufficient:
```suggestion
```
BTW the plots look very nice :+1:

src/util/plotting.jl
```diff
@@ -0,0 +1,38 @@
...
+            markershape --> :circle
+            markerstrokewidth --> 0.0
+            markersize --> 0.5
+            markeralpha --> 0.3
+            seriescolor --> "red"
+            label := ""
+
+            x, samples
+        end
+    end
```
devmotion, line 37: Maybe it would be better if one could plot samples separately, maybe using a `@userrecipe`? It seems a bit difficult/impossible to configure the plot of the GP and the samples manually in a different way.

sharanry: @devmotion are you referring to `@userplot`? I couldn't find anything called `@userrecipe`.

devmotion: Oh yes, sorry

devmotion: The main motivation to separate these two parts would be that then you can style both the samples and the standard plot of the GP independently. I guess that's not possible with the current version since, e.g., you can't just change the linewidth or color of the samples.

sharanry: `@userplot` seems to allow only args and not kwargs. So do you suggest we should give the rest of the arguments as a `Dict()`

sharanry: I am afraid this would become really messy. What if we just add all the style options of samples as kwargs to the current implementation? If sample=0 (which is the default), no samples will be shown.

devmotion: No, I think the plot recipe for a GP should just plot the GP, and there should be a separate plot recipe (maybe implemented with `@userplot`) that allows to plot samples of a GP, e.g., by calling `plotsamples(::AbstractGP, nsamples)` or whatever seems appropriate. AFAIK this can be achieved by defining `@userplot PlotSamples` and then implementing `@recipe function f(p::PlotSamples) ...`  and extracting the GP and the number of samples from `p` in the function body.

willtebbutt: I'm pro- whatever you two think is best here. I still don't really understand how plotting recipes work 🤷‍♂️

devmotion: I'm still confused every time I read through the documentation, so I'm not sure if I should give any recommendations on how to implement it properly :smile: I just think it would be easier to customize and work with if the sample part would be separate.

sharanry: > No, I think the plot recipe for a GP should just plot the GP, and there should be a separate plot recipe (maybe implemented with `@userplot`) that allows to plot samples of a GP, e.g., by calling `plotsamples(::AbstractGP, nsamples)` or whatever seems appropriate. AFAIK this can be achieved by defining `@userplot PlotSamples` and then implementing `@recipe function f(p::PlotSamples) ...` and extracting the GP and the number of samples from `p` in the function body.

Problem with plotting samples using `@userplot` is that we wouldn't be able to customize style options using kwargs. There would be no added flexibility when compared to current implementation.

devmotion: You can use keyword arguments with `@userplot`, so you can use `plotsamples(mygp, 20; linewidth = 30, mycustomkw = whatever)`. The macro just creates a type which the Plots machinery can use for dispatch, but it allows to use all the functionality available in regular `plot` calls: https://github.com/JuliaPlots/RecipesBase.jl/blob/cc3b95c95e0ac3cee7ab941ea297fdf453e76805/src/RecipesBase.jl#L339-L366

willtebbutt:
It would be really nice to write some tests for this so that we can at least guarantee that things will plot. Following up on a chat in the plots-bridged slack channel, apparently:

> Run the apply_recipe function and inspect the returned dict. No need to go any further unless you also want to check that changes to Plots will break functionality
To be clear, the recipe macro defines a new method for apply_recipe
I think maybe measurements.jl does this?

Again, I'm not really sure what the score is here, but I feel like we're going to need to add more plotting recipes moving forwards, so maybe this is a good thing to figure out?

src/util/plotting.jl
```diff
@@ -0,0 +1,64 @@
...
+        linealpha=0.2,
+        markershape=:circle,
+        markerstrokewidth=0.0,
+        markersize=0.5,
+        markeralpha=0.3,
+        seriescolor="red"
+    )
+"""
+sampleplot
+
```
devmotion, line 35: The docstring should be moved to `@userplot SamplePlot` (and probably modified, I don't think you want to use `f` there, and maybe it's better to have a section for keyword arguments instead of listing them in the function signature, see https://docs.julialang.org/en/latest/manual/documentation/):
```suggestion
```
As you can see in https://github.com/JuliaPlots/RecipesBase.jl/blob/cc3b95c95e0ac3cee7ab941ea297fdf453e76805/src/RecipesBase.jl#L356-L358, RecipesBase will add the docstring automatically to the definitions of `sampleplot!` and `sampleplot`.

src/util/plotting.jl
```diff
@@ -0,0 +1,64 @@
+using RecipesBase
+
+@recipe f(gp::AbstractGP, x::Array) = gp(x)
+@recipe f(gp::AbstractGP, x::AbstractRange) = gp(x)
+@recipe f(gp::AbstractGP, xmin::Number, xmax::Number) = (gp, range(xmin, xmax, length=1000))
+@recipe function f(gp::FiniteGP; samples=0, sample_seriestype=:line)
```
devmotion, line 6: I guess the keyword arguments are not needed anymore:
```suggestion
@recipe function f(gp::FiniteGP)
```

src/util/plotting.jl
```diff
@@ -0,0 +1,64 @@
...
+@userplot SamplePlot
+@recipe function f(sp::SamplePlot;
+        samples = 1,
+        sample_seriestype = :line,
+        linealpha=0.2,
+        markershape=:circle,
+        markerstrokewidth=0.0,
+        markersize=0.5,
+        markeralpha=0.3,
+        seriescolor="red"
```
devmotion, line 45: These keyword arguments are not needed:
```suggestion
```
Instead you should set defaults in the implementation below.

src/util/plotting.jl
```diff
@@ -0,0 +1,64 @@
...
+        samples = rand(f(x, 1e-9), samples)
+
+        seriestype --> sample_seriestype
+        linealpha := linealpha
+        markershape --> markershape
+        markerstrokewidth --> markerstrokewidth
+        markersize --> markersize
+        markeralpha --> markeralpha
+        seriescolor --> seriescolor
+        label := ""
```
devmotion, line 59: As mentioned above, defaults should be provided in the series definition:
```suggestion
        seriestype --> :line
        linealpha --> 0.2
        markershape --> :circle
        markerstrokewidth --> 0.0
        markersize --> 0.5
        markeralpha --> 0.3
        seriescolor --> :red
        label --> ""
```

src/util/plotting.jl
sharanry: I did check for that. However, I found this to be the case.
```julia
help?> sampleplot
search: sampleplot sampleplot!

  No documentation found.

  AbstractGPs.sampleplot is a Function.

  # 1 method for generic function "sampleplot":
  [1] sampleplot(args...; kw...) in AbstractGPs at /home/sharan/.julia/packages/RecipesBase/jcXIg/src/RecipesBase.jl:356

```

src/util/plotting.jl
sharanry: Sorry, I don't seem to understand. Don't we intend to the give the user the flexibility to change the defaults? If we were to set the defaults in the series definition,  how would the user modify them when required?

devmotion: That's all handled by the macros in RecipesBase. If you specify plotting options by `setting --> defaultvalue` then a user can specify a different setting by using the keyword argument `setting = someothervalue`. However, if you write `setting := value` then users can't specify a different value and it's fixed to `value`.

src/util/plotting.jl
devmotion: According to the implementation, writing
```julia
"""
My cool docstring.
"""
@userplot SamplePlot
```
should work. Did you try exactly this?

Project.toml
```diff
@@ -8,15 +8,15 @@ Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
...
```
devmotion, line 11: This issue is not resolved yet, Plots should only be a test dependency.

sharanry: Without importing `Plots`, `sampleplot` seems to give an error.
```julia
julia> sampleplot(gp; samples=10)
ERROR: MethodError: no method matching plot(::AbstractGPs.SamplePlot; samples=10)
Stacktrace:
 [1] sampleplot(::AbstractGPs.FiniteGP{GP{AbstractGPs.ZeroMean{Float64},SqExponentialKernel},Array{Float64,1},Diagonal{Float64,FillArrays.Fill{Float64,1,Tuple{Base.OneTo{Int64}}}}}; kw::Base.Iterators.Pairs{Symbol,Int64,Tuple{Symbol},NamedTuple{(:samples,),Tuple{Int64}}}) at /home/sharan/.julia/packages/RecipesBase/jcXIg/src/RecipesBase.jl:356
 [2] top-level scope at REPL[3]:1
 [3] eval(::Module, ::Any) at ./boot.jl:331
 [4] eval_user_input(::Any, ::REPL.REPLBackend) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.4/REPL/src/REPL.jl:86
 [5] run_backend(::REPL.REPLBackend) at /home/sharan/.julia/packages/Revise/C272c/src/Revise.jl:1075
 [6] top-level scope at REPL[1]:0

```

devmotion: You need Plots but only in the tests or if you want to use actually perform plotting. I.e., you have to write
```julia
julia> using AbstractGPs, Plots

julia> sampleplot(gp; samples = 10)
```
Users should not be forced to install or load Plots if they don't intend to plot anything.

src/util/plotting.jl
sharanry: I misunderstood you earlier. Thanks! It works.

Project.toml
sharanry: Okay seems fair. :) I will add the requirement of `Plots` in the docstring?

src/util/plotting.jl
```diff
@@ -0,0 +1,52 @@
...
+f = GP(SqExponentialKernel())
+sampleplot(f(rand(10), 10; markersize=5)
+```
+The given example plots 10 samples from the given `FiniteGP`. The `markersize` is modified from default of 0.5 to 10.
+"""
+@userplot SamplePlot
+@recipe function f(sp::SamplePlot)
+    x = sp.args[1].x
+    f = sp.args[1].f
+    samples = sp.args[2]
```
devmotion, line 38: I suggest to not use `samples` for both the number of samples and the actual samples.

sharanry:
@devmotion @willtebbutt Can we merge this?

src/util/plotting.jl
```diff
@@ -0,0 +1,52 @@
...
+
+Plot samples from the given `FiniteGP`. Make sure to run `using Plots` before using this function. 
+
+# Example
+```julia
+using Plots
+f = GP(SqExponentialKernel())
+sampleplot(f(rand(10), 10; markersize=5)
+```
+The given example plots 10 samples from the given `FiniteGP`. The `markersize` is modified from default of 0.5 to 10.
```
devmotion, line 32:
```suggestion
The given example plots 10 samples from the given `FiniteGP`. The `markersize` is modified from default of 0.5 to 5.
```

Project.toml
```diff
@@ -9,21 +9,21 @@ FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
...
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+RecipesBase = "3cdcf5f2-1ef4-517c-9805-6587b60abb01"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [compat]
 Distributions = "0.19, 0.20, 0.21, 0.22, 0.23"
 FillArrays = "0.7, 0.8"
-FiniteDifferences = "0.9.6"
 KernelFunctions = "0.4"
-Zygote = "0.4.6"
```
willtebbutt, line 19: These shouldn't have been dropped

src/util/plotting.jl
```diff
@@ -0,0 +1,52 @@
+using RecipesBase
+
+@recipe f(gp::AbstractGP, x::Array) = gp(x)
+@recipe f(gp::AbstractGP, x::AbstractRange) = gp(x)
+@recipe f(gp::AbstractGP, xmin::Number, xmax::Number) = (gp, range(xmin, xmax, length=1000))
```
willtebbutt, line 5:
```suggestion
@recipe f(gp::AbstractGP, xmin::Real, xmax::Real) = (gp, range(xmin, xmax, length=1000))
```
We probably don't want to allow complex numbers here :)

src/util/plotting.jl
```diff
@@ -0,0 +1,52 @@
+using RecipesBase
+
+@recipe f(gp::AbstractGP, x::Array) = gp(x)
```
willtebbutt, line 3:
```suggestion
@recipe f(gp::AbstractGP, x::AbstractVector) = gp(x)
```

src/util/plotting.jl
```diff
@@ -0,0 +1,52 @@
...
+Plot samples from the given `FiniteGP`. Make sure to run `using Plots` before using this function. 
+
+# Example
+```julia
+using Plots
+f = GP(SqExponentialKernel())
+sampleplot(f(rand(10), 10; markersize=5)
+```
+The given example plots 10 samples from the given `FiniteGP`. The `markersize` is modified from default of 0.5 to 5.
+"""
```
willtebbutt, line 33:
```suggestion
"""
    sampleplot(GP::FiniteGP, samples)

Plot samples from the given `FiniteGP`. Make sure to run `using Plots` before using this
function. 

# Example
```julia
using Plots
f = GP(SqExponentialKernel())
sampleplot(f(rand(10), 10; markersize=5)
```
The given example plots 10 samples from the given `FiniteGP`. The `markersize` is modified
from default of 0.5 to 5.
"""
```

File: test/runtests.jl [hunk 14]
```diff
@@ -3,13 +3,15 @@ using Documenter
...
 using Random
+using Plots
 using Test
 
 include("test_util.jl")
 
 @testset "AbstractGPs" begin
     @testset "util" begin
         include(joinpath("util", "common_covmat_ops.jl"))
+        include(joinpath("util", "plotting.jl"))
```
willtebbutt, line 14: This should be included after the tests for the rest of the package, since the plotting functionality won't work unless the rest of the package works. Just makes most sense to my mind to execute the tests for the package in the order that's given by the dependency structure -- children first, parents later.

Project.toml
sharanry: Thanks! Not sure how that happened.

sharanry:
@willtebbutt Could you take a second look? I have addressed your suggestions.

Project.toml
```diff
@@ -9,21 +9,24 @@ FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
 KernelFunctions = "ec8451be-7e33-11e9-00cf-bbf324bd1392"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+RecipesBase = "3cdcf5f2-1ef4-517c-9805-6587b60abb01"
```
willtebbutt, line 12: We need a compat bound for this.

devmotion:
We should bound `RecipesBase` but not `Plots` since the former is a dependency whereas the latter is not.

sharanry:
@willtebbutt If this is fine, could you approve this? Can't merge without an approval.

willtebbutt:
> We should bound RecipesBase but not Plots since the former is a dependency whereas the latter is not.

@devmotion is this true? It's a test-dependency according to our Project.toml, so should we not be requiring a compatible version?

devmotion:
> It's a test-dependency according to our Project.toml, so should we not be requiring a compatible version?

IMO ideally we should but it just doesn't work nicely with the old way of defining test dependencies in `Project.toml`: Pkg will remove these bounds every time you add or remove a dependency with `Pkg.add` or `Pkg.remove`, and IIRC CompatHelper won't open PRs for updating the compat bounds since Plots is only a test dependency. This can be handled in a much nicer way in Julia >= 1.2 by specifying the test dependencies in `test/Project.toml` instead in the usual way. IIRC CompatHelper will open PRs for Project.toml in subdirectories as well, so one should also be notified about updates of the dependencies.

devmotion:
Since this package has Julia 1.3 as lower dependency, we maybe should just remove the test dependencies from `Project.toml` and add a `test/Project.toml` instead.

willtebbutt:
> Since this package has Julia 1.3 as lower dependency, we maybe should just remove the test dependencies from Project.toml and add a test/Project.toml instead.

Yes, this is infinitely better. We should just do this.

devmotion:
BTW CompatHelper requires some additional configuration in that case:
https://github.com/bcbi/CompatHelper.jl#21-packages-in-subdirectories

sharanry:
@willtebbutt @devmotion I have created `test/Project.toml` and made changes to the config of `CompatHelper`. Could you check this once? I am not sure if I got the config right.

.gitignore
```diff
@@ -3,6 +3,7 @@
 *.jl.mem
 .DS_Store
 /Manifest.toml
+test/Manifest.toml
```
devmotion, line 6: This will ignore any file of name `Manifest.toml` that is in a subfolder `test` (not necessarily a subfolder of the root directory). If you want to ignore specifically the one in the test folder, you should write
```suggestion
/test/Manifest.toml
```
Alternatively, you could just change the line above to
```suggestion
Manifest.toml
```
to ignore all Manifest.toml files in any subfolder.

sharanry:
> Also, I would imagine that the `[extras]` block in `Project.toml` can be removed now.

I seem to get this error when I remove the `[extras]` section.
```julia
(@v1.4) pkg> activate .
ERROR: Compat `FiniteDifferences` not listed in `deps` or `extras` section.
```
I guess this is because `FiniteDifferences` is still in `Manifest.toml`. Would regenerating `Manifest.toml` fix this? If so, any idea how? I couldn't find anything when I searched for it.

sharanry:
I seem to get the same error even on removing `Manifest.toml`. Not sure why.

```julia
(AbstractGPs.jl) pkg> instantiate 
ERROR: Compat `FiniteDifferences` not listed in `deps` or `extras` section.
```

sharanry:
Regarding notebooks. I think they are the only way to retain figures. If large diff is a an issue, we will probably have to stick to `.jl` files.

willtebbutt:
Oh, I would guess that's just because there's still a `FiniteDifferences` compat bound in the main `Project.toml` file?

sharanry:
@willtebbutt Can we merge this if everything is fine? #10 is reliant on plotting utils.

willtebbutt:
@devmotion I'm happy if you are.

test/Project.toml
```diff
@@ -0,0 +1,23 @@
...
+FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
+FiniteDifferences = "26cc04aa-876d-5657-8c51-4c34ba976000"
+KernelFunctions = "ec8451be-7e33-11e9-00cf-bbf324bd1392"
+LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
+Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+RecipesBase = "3cdcf5f2-1ef4-517c-9805-6587b60abb01"
+Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
```
devmotion, line 13: Are all these dependencies needed? One only has to state the packages that are imported in the tests, so, e.g., I'm pretty sure RecipesBase could be removed (it's not imported explicitly and reexported by Plots). I guess FillArrays could be removed as well.

BTW that would be easier to see if all dependencies were imported in `test/runtest.jl`, currently some packages are loaded in `test/test_utils.jl` as well.

test/Project.toml
```diff
@@ -0,0 +1,23 @@
...
+
+[compat]
+Distributions = "0.19, 0.20, 0.21, 0.22, 0.23"
+FillArrays = "0.7, 0.8"
+FiniteDifferences = "0.9.6"
+KernelFunctions = "0.4"
+Plots = "1"
+RecipesBase = "1"
+Zygote = "0.4.6"
+julia = "1.3"
```
devmotion, line 23: This is missing a bound on Documenter. Also bounds for unneeded dependencies should be removed.

sharanry:
@devmotion Thanks for the review! I think I have addressed the issues you mentioned. Could you approve it so I can merge this?

PR #10, opened 2020-05-31
Add examples

sharanry:
This PR proposes to add a examples of
- Approximate inference of kernel parameters using `EllipticalSliceSampling.jl`
- Approximate inference of kernel parameters using `AdvancedHMC.jl`
- Approximate inference of kernel parameters and psuedo-points for Sparse GPs using Variational Inference.


examples/EllipticalSliceSampling.jl
```diff
@@ -0,0 +1,118 @@
...
+    kernel = ScaledKernel(transform(Matern52Kernel(), ScaleTransform(exp_params[1])), exp_params[2])
+    f = GP(kernel)
+    fx = f(x, 0.1)
+    p_fx = posterior(fx, y)
+    return logpdf(p_fx(x, 0.1), y)
+end
+
+# We define a Gaussian prior over the joint distribution on kernel parameters space. Since we have only two parameters, we define a multi-variate Gaussian of dimension two.
+
+prior = MvNormal(zeros(2), ones(2) .* 1)
```
devmotion, line 79:
```suggestion
prior = MvNormal(2, 1)
```
(see https://github.com/JuliaStats/Distributions.jl/blob/feae3ba9ada360d2980c901e7c737ca02bda7a31/src/multivariate/mvnormal.jl#L221). Another alternative would be
```suggestion
prior = MvNormal(ones(2))
```

examples/EllipticalSliceSampling.jl
```diff
@@ -0,0 +1,118 @@
...
+# Loading the necessary packages.
+
+using AbstractGPs, KernelFunctions, Plots, CSV, DataFrames
+include("utils.jl")
+
+# Loading [toy regression dataset](https://github.com/GPflow/docs/blob/master/doc/source/notebooks/basics/data/regression_1D.csv) taken from GPFlow examples.
+
+df = CSV.read("data/regression_1D.csv", header=false);
+x = df[:, 1];
+y = df[:, 2];
```
devmotion, line 27: Since the data is so small it feels like it would be easier to just copy the data here directly, instead of adding a CSV file and a dependency on CSV.

devmotion:
Maybe use Literate (or Weave alternatively), then you can commit just the script and generate the notebook when building the docs.

examples/EllipticalSliceSampling.jl
```diff
@@ -0,0 +1,118 @@
...
+
+# We define a function which returns log-likelihood of of data w.r.t an exact posterior with given set of kernel parameters.
+
+function logp(params)
+    exp_params = exp.(params)
+    kernel = ScaledKernel(transform(Matern52Kernel(), ScaleTransform(exp_params[1])), exp_params[2])
+    f = GP(kernel)
+    fx = f(x, 0.1)
+    p_fx = posterior(fx, y)
+    return logpdf(p_fx(x, 0.1), y)
```
willtebbutt, line 74: This should just be w.r.t. the prior `fx` rather than the posterior. The point of MCMC techniques is that you _don't_ need to evaluate the posterior density.

sharanry: Makes sense. Although, when I do this logpdf of the resulting model(logpdf = -13.230) isn't that great when compared to exact posterior (logpdf = -1.28). I guess this is expected especially considering that we are just doing an approximation with only 2 kernel parameters.

Is this fine or should I consider a kernel with more parameters?

devmotion:
> Is there any way of not including the 4K line diff that the notebook introduces? Just feels a bit unweildy.

[Literate](https://github.com/fredrikekre/Literate.jl/) would allow to only commit the Julia source files, and then notebooks and Markdown pages (for Documenter) can be created when generating the documentation.

sharanry:
@devmotion Thanks! I will try out `Literate.jl`. When I googled "Literate" earlier when you mentioned it, I found [zyedidia/Literate](https://github.com/zyedidia/Literate) which is a different project I guess. :sweat_smile:

docs/Project.toml
```diff
@@ -1,2 +1,4 @@
 [deps]
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+KernelFunctions = "ec8451be-7e33-11e9-00cf-bbf324bd1392"
+Literate = "98b081ad-f1c9-55d3-8b20-4c87d4299306"
```
devmotion, line 4: You might want to add the `docs` subfolder to CompatHelper.

examples/EllipticalSliceSampling.jl
```diff
@@ -0,0 +1,117 @@
...
+#     text_representation:
+#       extension: .jl
+#       format_name: light
+#       format_version: '1.5'
+#       jupytext_version: 1.4.2
+#   kernelspec:
+#     display_name: Julia 1.4.1
+#     language: julia
+#     name: julia-1.4
+# ---
```
devmotion, line 14: Is this still needed?

docs/src/generated/EllipticalSliceSampling.md
```diff
@@ -0,0 +1,162 @@
...
+sampleplot!(plt, p_fx(sort(x), 0.001), 100, alph=0.1)
+plt
+```
+
+-
+
+---
+
+*This page was generated using [Literate.jl](https://github.com/fredrikekre/Literate.jl).*
+
```
devmotion, line 162: It seems the file ended up in the PR although it should be ignored according to `.gitignore`.

sharanry: Thanks! They keep getting added automatically when I do `git add .`. Not sure why.

examples/EllipticalSliceSampling.jl
sharanry: It helps to keep this when I am editing with Jupyter notebook. The sync is done using jupytext using this metadata.I am pre-processing this out before generating the notebooks for the documentation. So, this won't appear there. 

Is it an issue if this stays here in `.jl` script? This would enable users to directly open this as a notebook.

devmotion: Can't you use `Literate.notebook` to generate the notebook and edit the script instead? Users could still open the notebook then.

sharanry: Yeah but AFAIK `Literate.notebook` doesn't allow syncing between the script and the notebook.

devmotion: No, it's just for generating the notebook from a Julia script.

sharanry: Yeah, so this extra metadata helps maintain two-way sync.

devmotion: Do we need two-way sync actually?

sharanry: I found a way to use jupyter notebooks directly with `.jl` files. I will remove the header.

sharanry:
@willtebbutt @devmotion Could you took a look at the VI example when every possible?

examples/EllipticalSliceSampling.jl
```diff
@@ -0,0 +1,114 @@
...
+
+# Loading [toy regression dataset](https://github.com/GPflow/docs/blob/master/doc/source/notebooks/basics/data/regression_1D.csv) taken from GPFlow examples.
+
+x = [0.8658165855998895, 0.6661700880180962, 0.8049218148148531, 0.7714303440386239, 0.14790478354654835, 0.8666105548197428, 0.007044577166530286, 0.026331737288148638, 0.17188596617099916, 0.8897812990554013, 0.24323574561119998, 0.028590102134105955];
+y = [1.5255314337144372, 3.6434202968230003, 3.010885733911661, 3.774442382979625, 3.3687639483798324, 1.5506452040608503, 3.790447985799683, 3.8689707574953, 3.4933565751758713, 1.4284538820635841, 3.8715350915692364, 3.7045949061144983];
+scatter(x, y, xlabel="x", ylabel="y")
+
+# Making a custom kernel with two parameters.
+
+k = ScaledKernel(transform(Matern52Kernel(), ScaleTransform(1.0)), 1.0)
```
devmotion, line 15: Wouldn't it be more intuitive to just use
```julia
k = Matern52Kernel()
```
here since all the optimization performed later on is not done with `k` but another kernel object with the same structure? The parameters of `k` are never tuned or adjusted.

examples/EllipticalSliceSampling.jl
```diff
@@ -0,0 +1,99 @@
...
+logpdf(p_fx(x), y)
+
+# Plotting the posterior `p_fx` along with the data points.
+
+plt = scatter(x, y, label = "Data")
+plot!(plt, p_fx, 0:0.001:1, label="Posterior")
+
+# # Elliptical Slice Sampler
+
+# Previously, we computed the the exact posterior GP without tuning the kernel parameters and achieved a loglikelihood on exact posterior of $-1.285$. We now attempt get a better posterior by tuning for kernel parameters using Elliptical Slice Sampler provided by [EllipticalSliceSampling.jl](https://github.com/TuringLang/EllipticalSliceSampling.jl/) instead of computing the exact posterior.
```
devmotion, line 44: IMO this explanation/motivation is a bit weird. AFAICT, it's just that above you computed the GP posterior for a GP model with Matern52Kernel as kernel, whereas now you consider a GP model with two parameters for which you perform Bayesian inference by specifying a prior distribution. IMO it's not correct to talk about an "exact posterior" (for a given prior you get a posterior - if you can evaluate it, it's always exact for the given prior) a "better posterior" (again, it's just the posterior for a different prior - at most an approximation of this posterior could be "better" (in whatever sense)) or "tuning the kernel parameters" (you just perform Bayesian inference for the kernel parameters as well).

sharanry: Thanks for pointing this out! This explanation has lost its meaning after many iterations of changes. 

How about changing it to this? - 
"Previously, we computed the posterior for a GP model with kernel having no parameters. We will now use Elliptical Slice Sampling(ESS) to perform Bayesian inference over parameter of the changed kernel. We use the kernel parameters' posterior samples obtained using (ESS) to generate samples from the posterior on the GP model with kernel having parameters to infer on."

examples/EllipticalSliceSampling.jl
```diff
@@ -0,0 +1,99 @@
...
+
+# Instantiating a `FiniteGP`, a finite dimentional projection at the inputs of the dataset observed under Gaussian Noise with $\sigma = 0.001$ .
+
+fx = f(x, 0.001)
+
+# Data's log-likelihood w.r.t prior `GP`. 
+
+logpdf(fx, y)
+
+# Calculating the exact posterior with the given expected output `y` and `FiniteGP`. THe GP's kernel currently has fixed parameters.
```
willtebbutt, line 29:
```suggestion
# Calculating the exact posterior over `f` given `y`. The GP's kernel currently has some arbitrary fixed parameters. 
```

examples/EllipticalSliceSampling.jl
```diff
@@ -0,0 +1,99 @@
...
+
+# Data's log-likelihood w.r.t prior `GP`. 
+
+logpdf(fx, y)
+
+# Calculating the exact posterior with the given expected output `y` and `FiniteGP`. THe GP's kernel currently has fixed parameters. 
+
+p_fx = posterior(fx, y)
+
+# Data's log-likelihood w.r.t exact posterior `GP`. We see that it drastically increases.
```
willtebbutt, line 33:
```suggestion
# Data's log-probability under the posterior `GP`. We see that it drastically increases.
```

examples/EllipticalSliceSampling.jl
```diff
@@ -0,0 +1,99 @@
...
+
+logpdf(p_fx(x), y)
+
+# Plotting the posterior `p_fx` along with the data points.
+
+plt = scatter(x, y, label = "Data")
+plot!(plt, p_fx, 0:0.001:1, label="Posterior")
```
willtebbutt, line 40:
```suggestion
plt = scatter(x, y; label="Data")
plot!(plt, p_fx, 0:0.001:1; label="Posterior")
```

examples/EllipticalSliceSampling.jl
```diff
@@ -0,0 +1,99 @@
...
+
+# Calculating the exact posterior with the given expected output `y` and `FiniteGP`. THe GP's kernel currently has fixed parameters. 
+
+p_fx = posterior(fx, y)
+
+# Data's log-likelihood w.r.t exact posterior `GP`. We see that it drastically increases.
+
+logpdf(p_fx(x), y)
+
+# Plotting the posterior `p_fx` along with the data points.
```
willtebbutt, line 37:
```suggestion
# Plot the posterior `p_fx` along with the observations.
```

examples/EllipticalSliceSampling.jl
```diff
@@ -0,0 +1,99 @@
...
+
+# # Elliptical Slice Sampler
+
+# Previously, we computed the the exact posterior GP without tuning the kernel parameters and achieved a loglikelihood on exact posterior of $-1.285$. We now attempt get a better posterior by tuning for kernel parameters using Elliptical Slice Sampler provided by [EllipticalSliceSampling.jl](https://github.com/TuringLang/EllipticalSliceSampling.jl/) instead of computing the exact posterior.
+
+# We start of by loading necessary packages.
+
+using EllipticalSliceSampling, Distributions
+
+# We define a function which returns log-likelihood of of data w.r.t a GP with the given set of kernel parameters.
```
willtebbutt, line 50:
```suggestion
# We define a function which returns log-probability of the data under the GP / log-likelihood of the parameters of the GP.
```

examples/EllipticalSliceSampling.jl
```diff
@@ -0,0 +1,99 @@
...
+logp(rand(prior))
+
+# Sampling 2,000 samples using `ESS_mcmc` provided by `EllipticalSliceSampling.jl`. 
+
+samples = ESS_mcmc(prior, logp, 2_000);
+samples_mat = hcat(samples...)';
+
+# Plotting a histogram of the samples for the two parameters. The vertical line in each graph indicates the mean of the samples.
+
+plt = histogram(samples_mat, layout=2, labels= "Param")
```
willtebbutt, line 75:
```suggestion
plt = histogram(samples_mat; layout=2, labels="Param")
```

examples/EllipticalSliceSampling.jl
```diff
@@ -0,0 +1,99 @@
...
+vline!(plt, mean(samples_mat, dims=1), layout=2, label="Mean")
```
willtebbutt, line 76:
```suggestion
vline!(plt, mean(samples_mat; dims=1); layout=2, label="Mean")
```

examples/EllipticalSliceSampling.jl
```diff
@@ -0,0 +1,99 @@
...
+# Sanity check for the defined `logp` function and `prior` distribution.
+
+logp(rand(prior))
+
+# Sampling 2,000 samples using `ESS_mcmc` provided by `EllipticalSliceSampling.jl`. 
+
+samples = ESS_mcmc(prior, logp, 2_000);
+samples_mat = hcat(samples...)';
+
+# Plotting a histogram of the samples for the two parameters. The vertical line in each graph indicates the mean of the samples.
```
willtebbutt, line 73: It would be good to add some interpretation / analysis here.

examples/EllipticalSliceSampling.jl
```diff
@@ -0,0 +1,99 @@
...
+
+plt = histogram(samples_mat, layout=2, labels= "Param")
+vline!(plt, mean(samples_mat, dims=1), layout=2, label="Mean")
+
+# Mean of samples of both the parameters.
+
+mean_params = mean(samples_mat, dims=1)
```
willtebbutt, line 80: Do we need this, given that it's computed above?

examples/EllipticalSliceSampling.jl
```diff
@@ -0,0 +1,99 @@
...
+
+# Conditional log-probability of GP with kernel's parameters tuned using ESS. We can observe that there is significant improvement over exact posterior with default kernel parameters.
```
willtebbutt, line 82: I've made a suggestion here. I wonder whether this is the right thing to do here really. Is there a particular reason you chose the posterior mean of the parameters?

```suggestion
# Conditional log-probability of data with posterior mean kernel parameters sampled using ESS. We can observe that there is significant improvement over exact posterior with default kernel parameters. 
```

examples/EllipticalSliceSampling.jl
```diff
@@ -0,0 +1,99 @@
...
+mean_params = mean(samples_mat, dims=1)
+
+# Conditional log-probability of GP with kernel's parameters tuned using ESS. We can observe that there is significant improvement over exact posterior with default kernel parameters. 
+
+logp(mean_params)
+
+# Plotting sampled functions from posterior with tuned parameters
+
+
+plt = scatter(x, y, label="data")
```
willtebbutt, line 89:
```suggestion
plt = scatter(x, y; label="data")
```

examples/EllipticalSliceSampling.jl
```diff
@@ -0,0 +1,99 @@
+# # An example showcasing the usage of EllipticalSliceSampling.jl with AbstractGPs.jl
```
willtebbutt, line 1: One other thing -- this results in a really long title when one builds the docs, so it would be good to shorten this substantially to something like "Example: Approximate Inference"

examples/EllipticalSliceSampling.jl
willtebbutt: This is better. The third sentence still doesn't quite work though -- I think it can just be dropped tbh:

"Previously, we computed the posterior for a GP model with kernel having no parameters. We will now use Elliptical Slice Sampling(ESS) to perform Bayesian inference over the parameters of the length scale and variance parameters of the kernel."

examples/EllipticalSliceSampling.jl
sharanry: Since we will be showcasing usage of many different approximate inference schemes, maybe "Example: Approximate Inference using ESS" would be more appropriate?

examples/EllipticalSliceSampling.jl
sharanry: This is just to compare it with the prior log-likelihood. To show an improvement on average.

examples/EllipticalSliceSampling.jl
sharanry: Only think I can think of to add is how it has shifted from the prior over the parameters. Do you have something else in mind?

examples/EllipticalSliceSampling.jl
sharanry: I have shifted this above.

examples/EllipticalSliceSampling.jl
willtebbutt: Actually, I'm not really sure what I wanted.

examples/EllipticalSliceSampling.jl
willtebbutt: Hmmm this isn't really the right way to do that. You should evaluate the log marginal likelihood for each sample and then average / look at the distribution over that quantity.

sharanry: I jus thought this would give a rough estimate. I realize this isn't proper now.  Is it right to say, what you suggest takes a expectation over log-likelihood w.r.t posterior distribution of the parameters?

willtebbutt: Yeah, that's what I was suggesting. I would assess this on held-out test data though, so that you're computing a Monte Carlo estimate of `log p(y_test | p_train)`.

sharanry: I have tried incorporating what you suggested. Could you check it once?

sharanry:
@willtebbutt Thanks for the review.
I have tried to incorporate what you had suggested. We now adhere to 92 char per line limit in examples as much as possible (few cases with long links is probably unavoidable). The VI example now plots exact and approximate posterior together.

willtebbutt:
@sharanry the docs build appears to be hanging. I have literally no idea why it would be doing that.

docs/make.jl
```diff
@@ -1,13 +1,36 @@
 using Documenter, AbstractGPs
 
+using Literate
+using Plots # to not capture precompilation output
+
+function preprocessing(str)
+    return replace(str, r"# ---[a-zA-Z0-9 #:,.'-_\n]*# ---\n" => ""; count=1)
+end
```
devmotion, line 8: I don't think this is needed anymore?
```suggestion
```

docs/make.jl
```diff
@@ -1,13 +1,36 @@
...
+
+Literate.markdown(
+    joinpath(@__DIR__, "..", 
+    "examples/EllipticalSliceSampling.jl"), 
+    joinpath(@__DIR__, "src/generated"); 
+    name = "EllipticalSliceSampling", 
+    preprocess=preprocessing
```
devmotion, line 15:
```suggestion
```

devmotion:
Maybe ESS gets stuck in an infinite loop, even though mathematically that shouldn't happen? Can you build the docs locally without any problems?

examples/EllipticalSliceSampling.jl
```diff
@@ -0,0 +1,135 @@
+# # Example: Approximate Inference using ESS
+
+# Loading the necessary packages.
+
+using AbstractGPs, KernelFunctions, Plots
```
devmotion, line 5: Maybe set a seed?

examples/EllipticalSliceSampling.jl
```diff
@@ -0,0 +1,135 @@
...
+
+function logp(params; x=x_train, y=y_train)
+    exp_params = exp.(params)
+    kernel = ScaledKernel(
+        transform(
+            Matern52Kernel(), 
+            ScaleTransform(exp_params[1])
+            ), 
+        exp_params[2]
+    )
```
devmotion, line 78: Maybe just
```suggestion
    kernel = ScaledKernel(
        transform(
            Matern52Kernel(), 
            ScaleTransform(exp(params[1])),
        ), 
        exp(params[2]),
    )
```
?

examples/EllipticalSliceSampling.jl
```diff
@@ -0,0 +1,135 @@
...
+prior = MvNormal(2, 1)
+
+# Sanity check for the defined `logp` function and `prior` distribution.
+
+logp(rand(prior))
+
+# Generate 2,000 samples using `ESS_mcmc` provided by `EllipticalSliceSampling.jl`. 
+
+samples = ESS_mcmc(prior, logp, 2_000);
+samples_mat = hcat(samples...)';
```
devmotion, line 97: In general, I think it's not a good idea to splat 2000 samples. Maybe rather use
```suggestion
samples_mat = reduce(hcat, samples)';
```

examples/EllipticalSliceSampling.jl
```diff
@@ -0,0 +1,135 @@
...
+# graph indicates the mean of the samples. 
+
+plt = histogram(samples_mat; layout=2, labels="Param")
+vline!(plt, mean_params; layout=2, label="Mean")
+
+# Average log-marginal-probability of data with posterior kernel parameter samples 
+# sampled using ESS. We can observe that there is significant improvement over 
+# exact posterior with default kernel parameters.
+
+mean([logp(param; x=x_test, y=y_test) for param in samples])
```
devmotion, line 113: One could even drop the array:
```suggestion
mean(logp(param; x=x_test, y=y_test) for param in samples)
```

examples/EllipticalSliceSampling.jl
```diff
@@ -0,0 +1,135 @@
...
+scatter!(plt, x_train, y_train; label="Test data")
+for params in eachrow(samples_mat[end-100:end,:])
+    exp_params = exp.(params)
+    opt_kernel = ScaledKernel(
+        transform(
+            Matern52Kernel(), 
+            ScaleTransform(exp_params[1])
+            ), 
+        exp_params[2]
+    )
```
devmotion, line 128: Maybe also just
```suggestion
    opt_kernel = ScaledKernel(
        transform(
            Matern52Kernel(), 
            ScaleTransform(exp(params[1])),
        ), 
        exp(params[2])
    )
```

examples/EllipticalSliceSampling.jl
```diff
@@ -0,0 +1,135 @@
...
+# exact posterior with default kernel parameters.
+
+mean([logp(param; x=x_test, y=y_test) for param in samples])
+
+# Plot sampled functions from posterior with tuned parameters
+
+
+plt = scatter(x_train, y_train; label="Train data")
+scatter!(plt, x_train, y_train; label="Test data")
+for params in eachrow(samples_mat[end-100:end,:])
```
devmotion, line 120: Maybe
```suggestion
for params in @view(samples[(end-100):end])
```
or
```suggestion
for i in 1_900:2_000
    params = samples[i]
```
is a bit simpler?

sharanry:
> Maybe ESS gets stuck in an infinite loop, even though mathematically that shouldn't happen? Can you build the docs locally without any problems?

This gets stuck at the same place even in a local build. I don't think the examples are the problem they run fine locally to generate corresponding `.md` files. The cause for the problem is Documenter's `makedocs` function.

willtebbutt:
@sharanry looks like this is nearly good to go. I think you'll be able to make the docs build if you remove the `Manifest.toml` from the `docs` dir. I think it's probably pinning you to an earlier version of `AbstractGPs`. In general, it just shouldn't be there I don't think.

sharanry:
> @sharanry looks like this is nearly good to go. I think you'll be able to make the docs build if you remove the `Manifest.toml` from the `docs` dir. I think it's probably pinning you to an earlier version of `AbstractGPs`. In general, it just shouldn't be there I don't think.

The reason `docs/Manifest.toml` was included was to force docs environment to use the the package from the `src` folder instead of using the latest release version. The docs environment is pinned to the package by specifying path [here](https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/blob/ddd33c13931abcea220749d331cce20c6a3fdb9b/docs/Manifest.toml#L5).

willtebbutt:
Oh interesting, I hadn't realised that was the case. Good to know. I guess something else must be causing the docs not to find `ReExport.jl`.

yebai:
What’s missing from this PR?

sharanry:
> What’s missing from this PR?

Nothing really. The documentation builds fine but there is a problem with deployment. 

I tried asking on the Julia's documentation channel for help but no one could figure out what the problem is. Everyone suspects that it is something to do with the deploy key or the secret key which is not a part of the PR. I have been stuck with this for few weeks now.

sharanry:
@willtebbutt This looks good to go now. I will merge it no one has any objections.

yebai:
Good work, @sharanry!

PR #16, opened 2020-06-25
Resolves #14

### LINKED ISSUE #14: ZeroMean{Float64} is not callable
So diligently using `ColVecs` :smile: I faced the following bug when trying to optimize kernel parameters. Here is the MWE
```julia
using AbstractGPs, KernelFunctions, Flux
X = rand(2, 10)
y = rand(10)
l = rand(2)
k = transform(SqExponentialKernel(), l)
Flux.gradient(Flux.params(k)) do
    m = GP(k)(KernelFunctions.ColVecs(X), 0.01)
    logpdf(m, y)
end
```
And get the error `LoadError: MethodError: objects of type AbstractGPs.ZeroMean{Float64} are not callable`

One additional note : `KernelFunctions` and `ColVecs/RowVecs` should be exported.

willtebbutt:
This will resolve the issue once a version of Flux has been released that supports the latest version of Zygote.


Project.toml
```diff
@@ -4,6 +4,7 @@ authors = ["willtebbutt <wt0881@my.bristol.ac.uk>"]
 version = "0.1.1"
 
 [deps]
+ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
```
sharanry, line 7: I think we should add compat entry for `ChainRulesCore.jl`.

test/Project.toml
```diff
@@ -1,4 +1,5 @@
 [deps]
+ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
```
sharanry, line 2: Same here, I think we should add a compat entry. I can't seem to add a code change suggestion.

test/gp/mean_functions.jl
```diff
@@ -1,23 +1,27 @@
...
-        x = randn(rng, N)
-        foo_mean = x->sum(abs2, x)
-        f = CustomMean(foo_mean)
-
-        @test map(f, x) == map(foo_mean, x)
-        # differentiable_mean_function_tests(f, randn(rng, N), x)
-    end
     @testset "ZeroMean" begin
-        rng, P, Q, D = MersenneTwister(123456), 3, 2, 4
+        P = 3
```
sharanry, line 3: Any reason why we don't want to set a random seed?

Project.toml
willtebbutt: Good catch.

test/Project.toml
willtebbutt: Another good catch.

> I can't seem to add a code change suggestion. 🤔

Hmm that is strange.

test/gp/mean_functions.jl
willtebbutt: Turns out that whenever you enter a new `@testset` the default seed is reset, so there's actually no need to seed inside a test set unless you're doing very particular things. I only became aware of this recently, hence the change.

sharanry: This is good to know. Thanks!

willtebbutt:
> Maybe also export KernelFunctions and ColVecs/RowVecs as @theogf suggested in #14,

Done as per request. @sharanry let me know if there's anything else.

theogf:
I am going to open a different issue about reexporting `KernelFunctions` as I guess it needs to be discusses

theogf:
Is it fixed with Flux 0.11 ?

willtebbutt:
@sharanry are you happy with this PR now?

theogf:
I think in your merge you removed the export for `ColVecs` and `RowVecs`

willtebbutt:
> I think in your merge you removed the export for ColVecs and RowVecs

Yeah, is that not handled by reexport now?

theogf:
Well KernelFunctions is not exporting ColVecs and RowVecs

willtebbutt:
Oh. That's unfortunate. Probably it should be. In any case, I'll add the exports back in.

src/gp/mean_function.jl
```diff
@@ -7,7 +7,12 @@ Returns `zero(T)` everywhere.
 """
 struct ZeroMean{T<:Real} <: MeanFunction end
 
-Base.map(::ZeroMean{T}, x::AbstractVector) where T = zeros(T, length(x))
+Base.map(::ZeroMean{T}, x::AbstractVector) where {T} = zeros(T, length(x))
+
+function ChainRulesCore.rrule(::typeof(Base.map), m::ZeroMean, x::AbstractVector)
+    map_ZeroMean_pullback(Δ) = (NO_FIELDS, NO_FIELDS, Zero())
+    return map(m, x), map_ZeroMean_pullback
+end
```
theogf, line 15: Could you also add `rrule` for the other `MeanFunction` types?

willtebbutt: I would prefer to handle that in a separate PR -- better just to get this in for now and raise an issue about the other ones.

willtebbutt:
> This might be a silly question but is there any particular reason why are choosing to make use of reverse mode AD as opposed to forward-mode AD?

Not silly at all. Reverse-mode is generally better for GP stuff because the objective function maps from a high(-ish)-dimensional parameter space to a scalar, the log marginal probability of the data given the parameters. This is the context in which reverse-mode works well. Conversely, in this situation forwards-mode requires time proportional to the number of parameters. So if you only care about a few hyperparameters, forwards-mode is a good option. Once you've got more than a handful though, reverse-mode will typically dominate.

PR #18, opened 2020-06-26
Stop exporting mean_and_cov_diag

### LINKED ISSUE #17: mean_and_cov_diag not defined for predictions with
With a given `m isa PosteriorGP`, `mean_and_diag_cov` does not work when calling it on `m(Xtest)`
MWE:
```julia
using AbstractGPs, KernelFunctions
x = rand(10)
y = rand(10)
k = SqExponentialKernel()
m = posterior(GP(k)(x, 0.01), y)
mean_and_cov_diag(m(x))
```

willtebbutt:
`mean_and_cov_diag` is currently exported, but it shouldn't be. The level at which users are intended to operate is the `FiniteGP` level, not the `AbstractGP` level. For example:
```julia
AbstractGPs.mean_and_cov_diag(f, x)
```
is not the intended use, rather
```julia
marginals(f(x))
```
is what is intended.

@theogf does this resolve #17? The README example actually already shows this use case correctly, but hopefully this will dissuade users against using this version.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "AbstractGPs"
 uuid = "99985d1d-32ba-4be9-9821-2ec096f28918"
 authors = ["willtebbutt <wt0881@my.bristol.ac.uk>"]
-version = "0.1.4"
+version = "0.2"
 
 [deps]
 Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "AbstractGPs"
 uuid = "99985d1d-32ba-4be9-9821-2ec096f28918"
 authors = ["willtebbutt <wt0881@my.bristol.ac.uk>"]
-version = "0.1.3"
+version = "0.1.4"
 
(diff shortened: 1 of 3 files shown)


theogf:
I got it that it should not be exported, but isn't it a bit inconsistent that it does not work for the case shown #17 but for all other cases ?

willtebbutt:
Ah, no, it actually never works for `FiniteGP`s. Only the binary version ever works, and it's never the intention that users use the binary versions of functions, rather they should construct a `FiniteGP` pass that in. I'm tempted to rename some functions in the `AbstractGP` interface so that their names don't clash with the things are that exported.

For example, I consider
```julia
mean_and_cov(f, x)
```
an internal method, whike
```julia
mean_and_cov(f(x))
```
is the right thing for users to call. Probably I should rename some stuff to make this clearer.

The rationale is purely that there should be exactly one way of doing things whenever possible, and the latter is more general than the former.

theogf:
I think you need then to export `mean_and_cov_diag` in the test as wel

willtebbutt:
@theogf ready for review

PR #20, opened 2020-06-26
Update README

sharanry:
Updating README was overdue. I have tried to demonstrate all important features. It is pretty long and filled with code right now. Maybe the main page of the documentation might be a better place to add all this. I would be happy to hear you thoughts on how/where to demonstrate the capabilities of this package.

@willtebbutt @theogf @devmotion


devmotion:
I like these GP plots :smiley: But I agree, it's probably too much code and the details should be moved to the documentation. IMO it would be nice to provide some illustrations and textual summary of the features of the package in the README but move (most of) the more concrete examples to the documentation.

willtebbutt:
I agree with @devmotion . I'm really pro- having a very basic example with working code in the README that a prospective user can run, but we don't need all of the examples there. The docs are the right place for it.

yebai:
How about merging this PR for now, and gradually improve the docs? We can simplify the README file when the docs are ready.

README.md
```diff
@@ -9,46 +9,108 @@
 
 AbstractGPs.jl is a package that defines a low-level API for working with Gaussian processes (GPs), and basic functionality for working with them in the simplest cases. As such it is aimed more at developers and researchers who are interested in using it as a building block than end-users of GPs.
 
+![GP](gp.gif)
```
yebai, line 12: Would be helpful to have a section on plotting too.

theogf:
@yebai We have now this project to make common docs for the whole JuliaGaussianProcesses org, see #24

yebai:
It might be helpful to have some doc somewhere before the docs site is online, which might take sometime.

README.md
sharanry: Sorry for the late reply! I will add this in the next PR for readme.

yebai:
Sorry, guys - I merged this PR without asking first. I didn't think this is going to cause any harm since it is only docs improvements. I agree with @theogf that these docs should live in one place in the future.

PR #22, opened 2020-06-26
Reexport KernelFunctions.jl

### LINKED ISSUE #19: Should AbstractGPs reexport KernelFunctions?
I think it would make sense for `AbstractGPs` to reexport `KernelFunctions.jl` via [Reexport.jl](https://github.com/simonster/Reexport.jl). Or at least reexport all the `Kernel` (by iteratively going over the subtypes).
Connected to a demand in #17

theogf:
Fixes #19
One possible issue is that all the kernel functions (`kernelmatrix` etc) are exported as well, which is not really needed. 
@devmotion @willtebbutt I could also refine the PR to only reexport all the `Kernel` constructors.


willtebbutt:
I'm happy with this PR as is. I think the question of what is exported from KF is a different matter, but it's reasonable to wonder whether we need to be exporting anything other than the kernels from KF at all.

theogf:
Can someone reapprove the PR?

Issue #9, opened 2020-05-20
Multi-output GPs


willtebbutt:
There are a couple of options here.

## Multi-output GPs are just GPs

A multioutput GP can just be viewed as a single output GP on an extended input domain. Specifically, suppose we're interested in modelling functions $\mathcal{X} \to \mathbb{R}^N$. You can always rephrase this as a function $\mathcal{X} \times \{1, ..., N\} \to \mathbb{R}$. Consequently you can always re-write a multioutput GP as a single-output GP on an extended input domain.

## Multi-output GPs are there own special thing

The other approach is to introduce a custom object here that really treats a multi-output GP as a separate thing, and introduce new API features / conventions to handle them.

## Pros and Cons

There are a couple of advantages to the first approach: namely that there's no separate notion of a multi-output GP floating around the code base, just extra abstractions (probably just an extra `AbstractVector` type and some specifically "multi-output" kernels). This also means that things like pseudo-point approximations will just work with the multi-output models and be non-mean-field by default.

That being said, the second approach makes the multi-output structure _really_ explicit, and it's often the case that you want to be able to get at this for computational reasons. However, as I mentioned we can just introduce some new abstractions for multi-output GPs if we go with the everything-is-just-a-GP approach, making this look slightly less like an advantage.

Overall I definitely favour the everything-is-a-GP approach because it fits nicely with our current way of doing things, and makes multi-output GPs trivial from an API perspective -- if you're comfortable with things like the `ColVecs` and `RowVecs` abstractions, multi-output GPs will be completely natural.

sharanry:
>A multioutput GP can just be viewed as a single output GP on an extended input domain. Specifically, suppose we're interested in modelling functions $\mathcal{X} \to \mathbb{R}^N$. You can always rephrase this as a function $\mathcal{X} \times {1, ..., N} \to \mathbb{R}$. Consequently you can always re-write a multioutput GP as a single-output GP on an extended input domain.

I am not sure how this is made possible. Are we assuming all multi-output functions we model are invertible in some sense and go on to the model the inverse of the function?

Few other ways we could design a multi-output GP.  
1. Have separate kernels for each output dimension and have combined kernel which would incorporate these separate kernels along with a co-variances between the two output dimensions. Refer to pg 7 in [1] OR have shared independent multi-output kernel as suggested in [2]. 
3. Shared pseudo-points for all dimensions OR separate set of pseudo points for each dimensions. Also discussed in [2].

[...]

PR #28, opened 2020-07-12
Add GP which models multi-dimensional output 

sharanry:
Addresses #9 

This PR hopes to introduce types which enable easy handling of multidimensional outputs.

TODO:

- [ ] Improve doctrings.
- [ ] Add `mean` and `cov` functions.
- [ ] Add example.


willtebbutt:
@sharanry and I discussed this separately. It probably makes sense to move most of the functionality from the PR to `KernelFunctions.jl`. I think that we should revisit how `AbstractGPs` should treat multi-output GPs, if it should do anything specialised at all, once the functionality has been added there.

sharanry:
Closed in favor of [KernelFunctions#138](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/138)

PR #30, opened 2020-07-25
Add MOGP example

sharanry:
Please find a simple example use of Independent Multi-Output Kernel to model Multi-output data. We use a synthetic regression dataset with 1-dim inputs and 2-dim outputs for ease of visualization.

Related PR: [KernelFunctions#138](https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/138)
Related Issue: #9 

Couple of improvements which I felt are required for seamless support of MOGPs.
- Plotting methods which support MOGPs. Edit: This is probably unnecessary as we hardly every "plot" results of multi-dim data unless its <= 3.
- Utility functions which allows automatic transformation forward and inverse transformation of outputs along with the inputs.

mogp

@willtebbutt @theogf @devmotion


theogf:
For multi-output with K >= 3, you could overlap all the outputs in a 2D plot.

sharanry:
> For multi-output with K >= 3, you could overlap all the outputs in a 2D plot.

What is `K` here?

theogf:
> > For multi-output with K >= 3, you could overlap all the outputs in a 2D plot.
> 
> What is `K` here?

Sorry, `K` is the dimension of the output

sharanry:
> Sorry, `K` is the dimension of the output
> I was refering to :
> 
> > This is probably unnecessary as we hardly every "plot" results of multi-dim data unless its <= 3.

Oh okay. I think I get what you mean. We can think of it like multiple 1D-1D functions.

yebai:
Perhaps consider adding a reference for the multiple output implementation in this PR.

sharanry:
> Perhaps consider adding a reference for the multiple output implementation in this PR.

@willtebbutt @yebai 
Would [A Framework forInterdomain and Multioutput Gaussian Processes](https://arxiv.org/pdf/2003.01115.pdf) be an appropriate reference?

willtebbutt:
Hmmm I'm not sure that that's an ideal reference. I would say that we really just need to reference any kernels that we implement. So if you're going to implement the SLFM we should cite that.

If we're have a related work discussion somewhere, we should probably discuss that paper though.

willtebbutt:
IMHO it would be most useful to have some functionality to multi-output time series problems. So you could imagine that `plot` is set up to detect the `MOInputs` type, and plot multiple panels accordingly or something, rather than doing a 3D plot. While the 3D plot is definitely cool, it's not usually what you're after (at least in my experience).

sharanry:
@theogf @willtebbutt I have tried to add the plotting utils you requested. 

`multidataplot`:


`multigpplot`:


This will work with arbitrary input and output dimensions.
Is this fine?

src/util/mogps.jl
```diff
@@ -0,0 +1,54 @@
...
+[`mo_inverse_transform`](@ref). 
+
+...
+# Arguments
+- `x::AbstractVector`: the inputs.
+- `y::AbstractVector`: the outputs.
+- `out_dim::Int`: the output dimension.
+...
+"""
+function mo_transform(x::AbstractVector, y::AbstractVector, out_dim::Int)
```
willtebbutt, line 15: I'm a bit confused by this.

The `logpdf` requires a `FiniteGP` and an `AbstractVector{<:Real}` of things. So I think this is doing what I expected for `x`, but I was expecting it to work the other way around for `y` -- it would accept `Y`, a `Matrix{<:Real}`, and return a `Vector{<:Real}` `y` that's been appropriately reshaped from `Y`.

Could you maybe elaborate a bit on why you've done it this way?

sharanry: The assumptions I am making on the initial data `x` and `y` is that they are Vector of Vectors/Reals. Where each element of the Vector is the input/output for one observation. 
Ex, 
```julia
in_dim=2; out_dim=3;
N=10; # no. of data points
x = [rand(in_dim) for _ in 1:N]
y = [rand(out_dim) for _ in 1:N]
```

On transforming `x` with `X = MOInput(x, out_dim)` we get an `AbstractVector` of length `out_dim*N`. So the GP would expect the targets it receives to be of same length. We do this by essentially flattening `y`(Vector of `N` Vectors of length `out_dim` each) to get `Y = vcat(([yi[i] for yi in y] for i in 1:out_dim)...) `.

Am I making sense?

willtebbutt: Ohhh sorry, I see. I misunderstood what you're doing.

I'm definitely on board with `x`, so that's fine.

I think what you're doing with `y` also makes sense though. Could you add specialisations for when `y` is a `ColVecs` or `RowVecs`? You can achieve the same results just using a reshape, which will be way more performant in general. A lot of the time the natural format for multi-output data will be a `Matrix{<:Real}`, so it would be great to support that.

examples/MOGPs.jl
```diff
@@ -0,0 +1,56 @@
...
+plt = multidataplot(x, y, in_dim, out_dim, layout=(1,3))
+plot!(plt, size=(900, 300), label="")
+
+# Make inputs and outputs compatible with multi-output GPs.
+
+X, Y = mo_transform(x, y, 2);
+
+# Define a multi-output kernel which assumes the outputs are independent of each other.
+
+k = IndependentMOKernel(MaternKernel())
```
theogf, line 25: I would not use `MaternKernel()` as an example as it is one of the most unstable one. Maybe just replace it by `Matern32Kernel()` ?

src/util/mogps.jl
```diff
@@ -0,0 +1,91 @@
...
+# Arguments
+- `x`: the input `Vector`.
+- `y`: the output `Vector`.
+- `out_dim::Int`: the output dimension.
+...
+"""
+function mo_transform(x::AbstractVector, y::AbstractVector, out_dim::Int)
+
+    X = MOInput(x, out_dim)
+    Y = vcat(([yi[i] for yi in y] for i in 1:out_dim)...)
```
theogf, line 28: Can we construct a different type which does not allocate? In the same style as `ColVecs`, `MOInput` etc

src/util/mogps.jl
```diff
@@ -0,0 +1,91 @@
...
+- `x`: the input matrix of size `(in_dim, N)`.
+- `y`: the output matrix of size `(out_dim, N)`.
+...
+"""
+function mo_transform(x::AbstractMatrix, y::AbstractMatrix)
+    size(x, 2) == size(y, 2) || error("`x` and `y` are do not have compatible sizes.")
+    out_dim = size(y, 1)
+    X = MOInput(ColVecs(x), out_dim)
+    y_ = ColVecs(y)
+    Y = vcat(([yi[i] for yi in y_] for i in 1:out_dim)...)
```
theogf, line 46: Same here, why are we using `ColVecs` to return a `Matrix` in the end?

src/util/mogps.jl
devmotion: I'm wondering the same - this seems both a bit inefficient and ugly, maybe indicating that we're not using the correct data structure/format?

sharanry: We are not really allocating any extra space. This is just flattening the outputs using the same amount of space. 

Do you suggest we have something like `MOOutput` just for the sake of flattening? If we were to define such an type, I think KernelFunctions.jl might be a better place as `MOInput` is defined there.

theogf: I would strongly disagree. We can make them nicely compatible but a `MOOutput` would make no sense in `KernelFunctions` since it does not have any use there

src/util/mogps.jl
```diff
@@ -1,3 +1,36 @@
...
+"""
+struct MOutput{T<:AbstractVector} <: AbstractVector{Real}
+    x::T
+    out_dim::Integer
+end
+
+Base.length(out::MOutput) = out.out_dim * length(out.x)
+
+Base.size(out::MOutput, d) = d::Integer == 1 ? out.out_dim * size(out.x, 1) : 1 
+Base.size(out::MOutput) = (out.out_dim * size(out.x, 1),)
```
devmotion, line 14: I think since it's an AbstractVector and you define `length`, none of these are needed:
```suggestion
```

src/util/mogps.jl
```diff
@@ -1,3 +1,36 @@
+"""
+    MOutput(x::AbstractVector, out_dim::Integer)
+
+A data type to handle multi-dimensional outputs.
+"""
+struct MOutput{T<:AbstractVector} <: AbstractVector{Real}
+    x::T
+    out_dim::Integer
```
devmotion, line 8: Should avoid abstract fields:
```suggestion
    out_dim::Int
```
However, I'm wondering we actually need this - couldn't one just use `length(first(x))` in all places where it's needed?

src/util/mogps.jl
```diff
@@ -1,3 +1,36 @@
...
+end
+
+Base.length(out::MOutput) = out.out_dim * length(out.x)
+
+Base.size(out::MOutput, d) = d::Integer == 1 ? out.out_dim * size(out.x, 1) : 1 
+Base.size(out::MOutput) = (out.out_dim * size(out.x, 1),)
+
+Base.lastindex(out::MOutput) = length(out)
+Base.firstindex(out::MOutput) = 1
```
devmotion, line 17: I'm surprised that this is needed? I though this should be fine as well automatically.
```suggestion
```

src/util/mogps.jl
```diff
@@ -1,3 +1,36 @@
+"""
+    MOutput(x::AbstractVector, out_dim::Integer)
+
+A data type to handle multi-dimensional outputs.
+"""
+struct MOutput{T<:AbstractVector} <: AbstractVector{Real}
```
theogf, line 6: A few things. I don't think inheriting from `AbstractVector{Real}` will work. It would be better to be able to infer the `eltype` of `y`.

Could you create maybe multiple constructors? I suppose `out_dim` can be inferred in certain cases.
Also could you explicit more what the shape of `x` is? It is very unclear at first sight.

src/util/mogps.jl
```diff
@@ -1,3 +1,36 @@
...
+        if ind1 == 0 ind1 = len end
+        return out.x[ind1][ind2]
+    else
+        throw(BoundsError(string("Trying to access at ", ind)))
+    end
+end
+
+Base.iterate(out::MOutput) = (out[1], 1)
+Base.iterate(out::MOutput, state) = (state<length(out)) ? (out[state + 1], state + 1) : nothing
+
```
devmotion, line 33: I'm pretty sure there is no need to define `iterate` if `getindex` is implemented. But maybe I'm wrong?
```suggestion
```

src/util/mogps.jl
```diff
@@ -1,3 +1,36 @@
...
+function Base.getindex(out::MOutput, ind::Integer)
+    if ind > 0
+        len = length(out.x)
+        ind1 = ind % len
+        ind2 = (ind - 1) ÷ len + 1
+        if ind1 == 0 ind1 = len end
+        return out.x[ind1][ind2]
+    else
+        throw(BoundsError(string("Trying to access at ", ind)))
+    end
```
devmotion, line 28: Bounds checks should be implemented in a special way such that they can be eliminated with `@inbounds`:
https://docs.julialang.org/en/v1/devdocs/boundscheck/#Eliding-bounds-checks

src/util/mogps.jl
sharanry: I seem to get an error without them.
```julia
ERROR: LoadError: MethodError: no method matching size(::MOutput{Array{Array{Float64,1},1}})
```

devmotion: Ah seems to be the other way around: you don't need `length` if `size` is defined, and you have to define only `size(::MOutput)` but not the one with indices (see https://docs.julialang.org/en/v1/manual/interfaces/#man-interface-array).

devmotion:
```suggestion

Base.size(out::MOutput) = (out.out_dim * size(out.x, 1),)
```

src/util/mogps.jl
sharanry: We definitely can. I was unsure show expensive computing `length(first(x))` everytime would be.

src/util/mogps.jl
sharanry: Yup! These seem to have default fall-backs.

src/util/mogps.jl
sharanry: Yup even this seems to have default fallbacks.

src/util/mogps.jl
sharanry: @devmotion @theogf Do you suggest I remove it?

src/util/mogps.jl
```diff
@@ -0,0 +1,105 @@
...
+end
+
+Base.size(out::MOutput) = (out.out_dim * size(out.x, 1),)
+
+@inline function Base.getindex(out::MOutput, ind::Integer)
+    @boundscheck checkbounds(out, ind)
+    len = length(out.x)
+    ind1 = ind % len
+    ind2 = (ind - 1) ÷ len + 1
+    if ind1 == 0 ind1 = len end
```
devmotion, line 18: If I'm not mistaken, this can be written as
```suggestion
    ind1, ind2 = fldmod1(ind, length(out.x))
```

src/util/mogps.jl
sharanry: How would you suggest inferring the `eltype` of `x` and incorporating that into the type?

src/util/mogps.jl
sharanry: No they are not the same. All the tests fail if I replace it.

devmotion: Hmm I guess you should be able to use `fld1` and `mod1` instead of the current implementation but I assume I messed up the remainder etc. E.g., `ind1` is just `mod1(ind, len)`. So maybe it's just
 ```suggestion
    ind2, ind1 = fldmod1(ind, length(out.x))
```
?

Comment (sharanry) [reply src/util/mogps.jl]: This worked. Thanks!

File: src/util/mogps.jl [hunk 12]
```diff
@@ -0,0 +1,112 @@
+"""
+    MOutput(x::AbstractVector, out_dim::Integer)
+
+A data type to handle multi-dimensional outputs.
+"""
+struct MOutput{T<:AbstractVector} <: AbstractVector{eltype(eltype(T))} where T <: Real
```
sharanry, line 6: @theogf Is this what you had in mind?

theogf: So wait I want to be sure to understand what `MOutput` is.
We give it either a `Matrix` or a `Vector` of `Vector` of outputs, we store it as a `Vector` of `Vector` and then we want it to act like a long `Vector` of outputs right?
If so then it looks good.
This would be nonetheless nice to put all this description in the docs

devmotion, line 6: I would suggest
```suggestion
struct MOutput{T<:Real,X<:AbstractVector} <: AbstractVector{T}
```
here - does the current definition actually work? Also, I'm not sure if it is useful to wrap a matrix in this struct as well - e.g., the computation of the indices (in the flattened form) seems unnecessarily complicated for matrices since there you could just use the standard linear indexing anyways.

theogf: Haha yeah I am also surprised it just work.
In your suggested change, `T` is not inferred from the data...

devmotion: That would be part of the constructors, not of the type definition.

devmotion: If it would only wrap vectors of vectors, one could of course do stuff like
```suggestion
struct MOutput{T<:Real,X<:AbstractVector{<:AbstractVector{T}}} <: AbstractVector{T}
```
but the suggestion above would be more generally applicable, so I'm not sure if it's useful to be that specific here.

devmotion: BTW I'm not sure we should define wrappers for vectors of vectors from scratch here anyway if there are existing solutions such as https://github.com/SciML/RecursiveArrayTools.jl#vectorofarray.

theogf: But here that's not really what we want right? They just make a vector of vector looks like a matrix

devmotion: Oh yeah, I misremembered the VecOfArray thing, I somehow thought it's flattened as well.

theogf: I think https://github.com/jonniedie/ComponentArrays.jl could be a good candidate no? Some of it could actually be useful to recognize what output we are dealing with (named outputs would be amazing :D)

src/util/mogps.jl
```diff
@@ -1,22 +1,34 @@
...
     MOutput(x::AbstractMatrix)
+
+If the input is a `Matrix`, it expects the size of the input `x` to be `(out_dim, N)`
+where `out_dim` is the output dimension and `N` is the number of data-points.
 """
-MOutput(x::AbstractMatrix) = MOutput(ColVecs(x), size(x, 1))
+function MOutput(x::X) where X <: AbstractMatrix
+    cv = ColVecs(x)
+    return MOutput{eltype(X), typeof(cv)}(cv, size(x, 1))
+end
```
devmotion, line 20: Wouldn't it be more efficient to just return `vec(X)` in this case?
I don't see why we would want to wrap a Matrix inside of a ColVecs struct which is then wrapped inside of a MOutput object such that we can index it like a vector if we can just linearize the matrix in the first place?

sharanry: I thought the whole point was to avoid re-allotment of memory. `vec(x)` probably allocates memory?

devmotion: No, it doesn't, the docstring mentions e.g.:

> The resulting array shares the same underlying data as a, so it
  will only be mutable if a is mutable, in which case modifying one will also modify the other.

As an example, `vec` is used by NLsolve for all input data just to ensure that it can be accessed linearly.

theogf: TIL!

src/util/mogps.jl
```diff
@@ -14,10 +14,7 @@ end
 If the input is a `Matrix`, it expects the size of the input `x` to be `(out_dim, N)`
 where `out_dim` is the output dimension and `N` is the number of data-points.
 """
-function MOutput(x::X) where X <: AbstractMatrix
-    cv = ColVecs(x)
-    return MOutput{eltype(X), typeof(cv)}(cv, size(x, 1))
-end
+MOutput(x::AbstractMatrix) = vec(permutedims(x))
```
theogf, line 17: Maybe this is a stupid comment but couldn't be just expect the input to be `(N, out_dim)` instead? That's easier to create `[y1 y2]` and we avoid `permutedims`

devmotion: Maybe it would make sense to define this method only with an additional `obsdim` keyword argument for `AbstractMatrix` and implement optimized implementations for `RowVecs` and `ColVecs` (which make use of `vec`) to which this is dispatched?

I also don't like the fact that this method is called `MOutput` but does not create an `MOutput` object. Maybe instead define a method `moutput` that yields an `MOutput` instance for general vectors and a linear view of the underlying matrices for `RowVecs` and `ColVecs`?

sharanry: @devmotion 
> and implement optimized implementations for RowVecs and ColVecs (which make use of vec) to which this is dispatched?

I am not sure what you mean by this.

I have introduced `moutput`. Please take a look.

thomasgudjonwright (commented):
A quick question on the structure we want on the docs. Right now we have it that the 1-D regression example includes the basic fitting+plotting as well as some hyperparameter tuning on the same page. If we want to keep this structure the same maybe some kernel parameter tuning could be included here? Alternatively we could slightly restructure the docs to have different pages for basic examples and for the hyperparameter tuning (might be simpler).

thomasgudjonwright:
Also worth updating #95 as necessary as this goes through

willtebbutt:
Closing, as discussed in #95

PR #35, opened 2020-07-28
Remove dependency on AbstractGPs

devmotion:
See https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/pull/34#issuecomment-664841445.

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -1,5 +1,4 @@
 [deps]
-AbstractGPs = "99985d1d-32ba-4be9-9821-2ec096f28918"
 AdvancedHMC = "0bf59076-c3b1-5ca4-86bd-e02cd72cde3d"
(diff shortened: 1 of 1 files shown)


willtebbutt:
Why this rather than setting up the Manifest such that it always points to the right place?

devmotion:
I'd say there would be two "advantages":
- it would be consistent with `test/Project.toml` which should not/does not have to include a dependency on the tested package (see https://julialang.github.io/Pkg.jl/v1/creating-packages/#Test-specific-dependencies-in-Julia-1.2-and-above-1)
- CompatHelper doesn't try to add compat entries for AbstractGPs (we always want to use the package in the parent directory, so it's not really useful)

devmotion:
The Manifest.toml is still be used though to make the docs reproducible (see also https://discourse.julialang.org/t/psa-use-a-project-for-building-your-docs/14974).

PR #39, opened 2020-08-05
Fix dependencies

sharanry:
#36 #37 #38 

- Remove KernelFunctions from docs and test dependency as it is being reexported
- Add AbstractGPs to docs Manifest


theogf:
Haha you got faster than me (sorry I merged into master but then I reverted the changes). Maybe just update the bounds of the main Project to 0.4, 0.5 as well

theogf:
I think you need to remove explicit use of KernelFunctions for tests and docs. (remove `using KernelFunctions`)

docs/Manifest.toml
```diff
@@ -1,5 +1,11 @@
 # This file is machine-generated - editing it directly is not advised
 
+[[AbstractGPs]]
+deps = ["ChainRulesCore", "Distributions", "FillArrays", "KernelFunctions", "LinearAlgebra", "Random", "RecipesBase", "Reexport", "Statistics"]
```
sharanry, line 4: Looks like none of these dependencies are appearing the manifest as `AbstractGPs` is not mentioned in `docs/Project.toml`.

sharanry: I don't think it is a good idea to add them manually. Should we consider adding `AbstractGPs` back to the docs dependencies?

theogf:
For what reason are we keeping the Manifest.toml file btw? I think there was a discussion about it at some point.

docs/Manifest.toml
theogf: For what reason are we keeping the Manifest.toml file btw? I think there was a discussion about it at some point.

sharanry: > For what reason are we keeping the Manifest.toml file btw? I think there was a discussion about it at some point.


To make the docs build use `AbstractGPs` from `src` folder instead of the latest release version. We make that happen [here](https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/blob/502887f95ee4953191db0725a0dc730bae0d4bc8/docs/Manifest.toml#L5).

willtebbutt: So that we can make it always point at the appropriate version of `AbstractGPs` via the `path` field in the `Manifest`. Otherwise the `docs` / `test`s can point at the wrong version of `AbstractGPs`.

PR #51, opened 2020-09-04
Passing noise for matrix case

theogf:
Following thread on Slack, it is not possible to pass a noise argument in FiniteGP when using a matrix, i.e. :
```
X = rand(3,1)
gp = GP(SqExponentialKernel())
FiniteGP(gp, X, 0.01)
```
Currently does not work. This PR aims at fixing this behaviour and add necessary tests to check it.

### DIFF
--- a/src/abstract_gp/finite_gp.jl
+++ b/src/abstract_gp/finite_gp.jl
@@ -16,19 +16,10 @@ end
 
 const default_σ² = 1e-18
 
-FiniteGP(f::AbstractGP, x::AbstractVector, σ²::Real) = FiniteGP(f, x, Fill(σ², length(x)))
-
-FiniteGP(f::AbstractGP, x::AbstractVector) = FiniteGP(f, x, default_σ²)
+FiniteGP(f::AbstractGP, x::AbstractVector, σ²::Real=default_σ²) = FiniteGP(f, x, Fill(σ², length(x)))
 
 function FiniteGP(
-    f::AbstractGP, X::AbstractMatrix;
-    obsdim::Int = KernelFunctions.defaultobs
-    )
-    return FiniteGP(f, X, default_σ²; obsdim=obsdim)
-end
-
-function FiniteGP(
-    f::AbstractGP, X::AbstractMatrix, σ²::Real;
+    f::AbstractGP, X::AbstractMatrix, σ²=default_σ²;
     obsdim::Int = KernelFunctions.defaultobs,
 )
     return FiniteGP(f, KernelFunctions.vec_of_vecs(X; obsdim=obsdim), σ²)
--- a/src/abstract_gp/finite_gp.jl
+++ b/src/abstract_gp/finite_gp.jl
@@ -14,16 +14,24 @@ function FiniteGP(f::AbstractGP, x::AbstractVector, σ²::AbstractVector{<:Real}
     return FiniteGP(f, x, Diagonal(σ²))
 end
 
+const default_σ² = 1e-18
+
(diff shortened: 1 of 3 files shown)


src/abstract_gp/finite_gp.jl
```diff
@@ -14,16 +14,24 @@ function FiniteGP(f::AbstractGP, x::AbstractVector, σ²::AbstractVector{<:Real}
     return FiniteGP(f, x, Diagonal(σ²))
 end
 
+const default_σ² = 1e-18
+
 FiniteGP(f::AbstractGP, x::AbstractVector, σ²::Real) = FiniteGP(f, x, Fill(σ², length(x)))
```
devmotion, line 19: Maybe just
```suggestion
FiniteGP(f::AbstractGP, x::AbstractVector, σ²::Real = default_σ²) = FiniteGP(f, x, Fill(σ², length(x)))
```
?

src/abstract_gp/finite_gp.jl
```diff
@@ -14,16 +14,24 @@ function FiniteGP(f::AbstractGP, x::AbstractVector, σ²::AbstractVector{<:Real}
...
+    f::AbstractGP, X::AbstractMatrix;
+    obsdim::Int = KernelFunctions.defaultobs
+    )
+    return FiniteGP(f, X, default_σ²; obsdim=obsdim)
+end
 
 function FiniteGP(
-    f::AbstractGP,
-    X::AbstractMatrix;
+    f::AbstractGP, X::AbstractMatrix, σ²::Real;
```
devmotion, line 31: Maybe also just use
```suggestion
    f::AbstractGP, X::AbstractMatrix, σ²::Real = default_σ²;
```
here and remove the block above?

theogf: I wanted to keep the style from before. @willtebbutt was there any reason to have it this way?

willtebbutt: Yeah, I would just use a default argument and retain a single method involving a matrix, rather than having two.

It would also make sense to drop the type constraint here, since the noise variance doesn't have to be `Real`.

src/abstract_gp/finite_gp.jl
willtebbutt: This would be good. Could drop a method, which is always nice.

src/abstract_gp/finite_gp.jl
```diff
@@ -16,19 +16,10 @@ end
...
-
-FiniteGP(f::AbstractGP, x::AbstractVector) = FiniteGP(f, x, default_σ²)
-
-function FiniteGP(
-    f::AbstractGP, X::AbstractMatrix;
-    obsdim::Int = KernelFunctions.defaultobs
-    )
-    return FiniteGP(f, X, default_σ²; obsdim=obsdim)
-end
+FiniteGP(f::AbstractGP, x::AbstractVector, σ²::Real=default_σ²) = FiniteGP(f, x, Fill(σ², length(x)))
```
willtebbutt, line 19:
```suggestion
function FiniteGP(f::AbstractGP, x::AbstractVector, σ²::Real=default_σ²)
    return FiniteGP(f, x, Fill(σ², length(x)))
end
```
92 char lim

willtebbutt (approved):
LGTM modulo style remark.

willtebbutt:
(And patch version bump 🙂 )

src/abstract_gp/finite_gp.jl
theogf: Haha I was kinda expecting this one. I was not sure if it was valid here.

Issue #49, opened 2020-08-24
Can't use AbstractGPs with Turing


luiarthur:
Hi, 

Not sure if I'm using AbstractGPs the way it was intended.

I am able to extract the logpdf from AbstractGPs and use it to increment the log target density in Turing. But I am wondering if there is a simpler way to use AbstractGPs with Turing. For example, I feel inclined to do the following:

```julia
using Turing, AbstractGPs, KernelFunctions
import Random

# Define a kernel.
sqexpkernel(alpha::Real, rho::Real) = alpha^2 * transform(SqExponentialKernel(), 1/(rho*sqrt(2)))

@model GPRegression(y, X) = begin
    # Priors.
    alpha ~ LogNormal(0.0, 0.1)
    rho ~ LogNormal(0.0, 1.0)
    sigma ~ LogNormal(0.0, 1.0)
    
    # Realized covariance function
    kernel = sqexpkernel(alpha, rho)
    f ~ GP(kernel)
    
    # Sampling Distribution.
    y ~ f(X, sigma)
end;

Random.seed!(0)
y = randn(3)
X = randn(3, 1)
m = GPRegression(y, X)
@time chain = sample(m, NUTS(1000, 0.8), 1000)
```

But this raises the error: 

```julia
ArgumentError: Right-hand side of a ~ must be subtype of Distribution or a vector of Distributions.
```

Any comments/help would be appreciated.

luiarthur:
Apparently, I'm a dummy... Apparently, this works. (Changed the line `f ~ GP(kernel)` to `f = GP(kernel)`.)

```julia
using Turing, AbstractGPs, KernelFunctions
import Random

# Define a kernel.
sqexpkernel(alpha::Real, rho::Real) = alpha^2 * transform(SqExponentialKernel(), 1/(rho*sqrt(2)))

@model GPRegression(y, X) = begin
    # Priors.
    alpha ~ LogNormal(0.0, 0.1)
    rho ~ LogNormal(0.0, 1.0)
    sigma ~ LogNormal(0.0, 1.0)
    
    # Realized covariance function
    kernel = sqexpkernel(alpha, rho)
    f = GP(kernel)  # use `=` instead of `~`
    
    # Sampling Distribution.
    y ~ f(X, sigma)
end;

Random.seed!(0)
y = randn(3)
X = randn(3, 1)
m = GPRegression(y, X)
@time chain = sample(m, NUTS(1000, 0.8), 1000)
```

PR #52, opened 2020-09-04
Renamed logpdf(gp, y) to _logpdf(gp, y) to respect Distributions.jl API and add Turing compatibility test

theogf:
- Changed `Distributions.logpdf` to `Distributions._logpdf` to respect `Distributions.jl` API and ensure compatibility with the whole ecosystem.
See https://juliastats.org/Distributions.jl/latest/extends/#Create-a-Multivariate-Distribution-1
- Added `loglikelihood` for set of obervations (as a matrix)
- Added tests for `loglikelihood`
- Added a simple Turing model to test compatibility with the Turing framework (solves #53)


src/abstract_gp/finite_gp.jl
```diff
@@ -189,34 +189,11 @@ Random.rand(f::FiniteGP, N::Int) = rand(Random.GLOBAL_RNG, f, N)
...
-true
-```
-"""
-function Distributions.logpdf(f::FiniteGP, y::AbstractVector{<:Real})
+function Distributions._logpdf(f::FiniteGP, y::AbstractVector{<:Real})
     return first(logpdf(f, reshape(y, :, 1)))
 end
 
-function Distributions.logpdf(f::FiniteGP, Y::AbstractMatrix{<:Real})
+function Distributions._logpdf(f::FiniteGP, Y::AbstractMatrix{<:Real})
```
devmotion, line 196: Actually here it should probably be
```suggestion
function Distributions.logpdf(f::FiniteGP, Y::AbstractMatrix{<:Real})
```
to ensure that it is called instead of the default implementation for matrices in https://github.com/JuliaStats/Distributions.jl/blob/82120fcb8b51e0e3f5789b7ae78a9f9aa8ad2375/src/multivariates.jl#L230-L235. Alternatively, one could implement `Distributions._logpdf!` for matrices.

devmotion:
I guess it would be useful to implement `Distributions.loglikelihood(gp::FiniteGP, Y::AbstractMatrix{<:Real})` as well to avoid calling `_logpdf(gp, y)` for each observation separately, as implemented by the default fallback in https://github.com/JuliaStats/Distributions.jl/blob/82120fcb8b51e0e3f5789b7ae78a9f9aa8ad2375/src/multivariates.jl#L260-L263.

src/abstract_gp/finite_gp.jl
```diff
@@ -189,40 +189,26 @@ Random.rand(f::FiniteGP, N::Int) = rand(Random.GLOBAL_RNG, f, N)
...
-function Distributions.logpdf(f::FiniteGP, y::AbstractVector{<:Real})
-    return first(logpdf(f, reshape(y, :, 1)))
+function Distributions._logpdf(f::FiniteGP, y::AbstractVector{<:Real})
+    m, C_mat = mean_and_cov(f)
+    C = cholesky(Symmetric(C_mat))
+    T = promote_type(eltype(m), eltype(C), eltype(y))
+    return -((length(y) * T(log(2π)) + logdet(C)) + diag_Xt_invA_X(C, y - m)) / 2
 end
 
+Distributions.loglikelihood(f::FiniteGP, Y::AbstractMatrix{<:Real}) = logpdf(f, Y)
```
devmotion, line 199: That is not correct, `loglikelihood` has to return a scalar. The default implementation in Distributions computes `sum(i -> logpdf(f, view(Y, :, i)), axes(f, 2))` but this seems inefficient due to the repeated Cholesky factorizations.

src/abstract_gp/finite_gp.jl
```diff
@@ -189,40 +189,26 @@ Random.rand(f::FiniteGP, N::Int) = rand(Random.GLOBAL_RNG, f, N)
...
     m, C_mat = mean_and_cov(f)
     C = cholesky(Symmetric(C_mat))
     T = promote_type(eltype(m), eltype(C), eltype(Y))
     return -((size(Y, 1) * T(log(2π)) + logdet(C)) .+ diag_Xt_invA_X(C, Y .- m)) ./ 2
 end
 
+function Distributions._logpdf!(r::AbstractArray, f::FiniteGP, Y::AbstractMatrix{<:Real})
+    r .= logpdf(f, Y)
+end
+
```
devmotion, line 211: I don't think this is needed (at least not right now) if `logpdf(f, ::AbstractMatrix)` is implemented. If we want to add it, then `logpdf(f, ::AbstractMatrix)` should probably be removed to achieve a truly in-place implementation.

src/abstract_gp/finite_gp.jl
theogf: Ah ok got it! Sorry I went too fast

theogf:
> @theogf could you explain what advantages one gains from implementing this bit of the interface in the particular manner than `Distributions` asks for, over what we currently do? Note that the same criticism could be made of `rand`: technically we should be implementing `_rand`.

Note that I did not changed to `logpdf!` but to `_logpdf`. I will remove the `logpdf!` one because as @devmotion pointed out it's pretty much useless now.
The issue was raised after @elizavetasemenova tried to run the `Turing` model from #49 with `AbstractGPs` which then spitted errors such as `_logpdf` (called by `loglikelihood` called by `Turing`) is not defined for GPs.
The advantage is that we are adapting to API requirements of Distributions.jl.
And indeed we should probably implement `_rand!` as well : should I take care of it in this PR ?

willtebbutt:
> The issue was raised after @elizavetasemenova tried to run the Turing model from #49 with AbstractGPs which then spitted errors such as _logpdf (called by loglikelihood called by Turing) is not defined for GPs.

This is interesting. I was not aware that Turing ever uses `loglikelihood` -- last I checked it uses `logpdf` everwhere. Could you point me towards the error.

> Note that I did not changed to logpdf! but to _logpdf. I will remove the logpdf! one because as @devmotion pointed out it's pretty much useless now.

Okay.

> The advantage is that we are adapting to API requirements of Distributions.jl.

This is only an advantage if not implementing things precisely in the way that they expect is causing issues.

> And indeed we should probably implement _rand! as well : should I take care of it in this PR ?

I am not particularly in favour of this. We have to have an implementation of `rand` for the sake of AD, so implementing `_rand!` will necessitate either copying code or making a kind-of-in-place implementation of `_rand!` that calls `rand` internally. Unless there's an actual use-case that we care about, I don't want to have to maintain the extra code / add the extra complexity to the code base.

src/abstract_gp/finite_gp.jl
```diff
@@ -189,40 +189,26 @@ Random.rand(f::FiniteGP, N::Int) = rand(Random.GLOBAL_RNG, f, N)
...
-
-julia> logpdf(f(x), y) isa Real
-true
-
-julia> Y = rand(f(x), 3);
-
-julia> logpdf(f(x), Y) isa AbstractVector{<:Real}
-true
-```
-"""
```
willtebbutt, line 214: I would actually really like to keep this around.

`logpdf` with `GPs` is a bit of a weird concept for users, so I think it warrants explaining what is meant by the function in the context of `AbstractGPs`.

Also, this includes a doctest, which I would prefer not to lose.

theogf: I agree this docs are useful but when would they appear though? Calling `?logpdf` would return the `Distributions.jl` one no?

devmotion:
> @devmotion I'm also confused as to what the need to implement loglikelihood -- it's got a somewhat different meaning to logpdf.
>
> edit: I can see the advantage of _logpdf (as opposed to _logpdf!) as it appears to provide dimensionality checking -- so this seems like a reasonable thing to do. I'm still confused as to why loglikelihood isn't just logpdf, with the convention that if you provide a matrix of observations, you return a vector of logpdfs.

If one implements `Distributions._logpdf`, Distributions provides a default implementation of `Distributions.loglikelihood` for single + multiple samples (i.e. vectors, matrices, and vectors of vectors). However, the default implementation will just call `_logpdf` iteratively for each observation in the matrix which is inefficient since the Cholesky decomposition is then performed multiple times.

Turing 0.14 uses `loglikelihood` in multiple places (e.g., https://github.com/TuringLang/DynamicPPL.jl/blob/7afd70ccab7bf76682c8b9ddb17ef4954fe56d06/src/context_implementations.jl#L449) instead of `sum(logpdf(d, x))` since in particular for multi- and matrixvariate distributions the former allows for more efficient implementations that do not have to allocate (as much) (see, e.g., https://github.com/TuringLang/DistributionsAD.jl/blob/f754a19f2f8c005898674417b3cc71347526eb31/src/multivariate.jl#L146).
Maybe therefore we might even want to add a more efficient implementation of `loglikelihood(f::FiniteGP, Y::AbstractMatrix{<:Real})` that does not fall back to `sum(logpdf(f, Y))`.

src/abstract_gp/finite_gp.jl
devmotion: Maybe it is possible to keep the docstring in the following way?
```julia
"""
    logpdf(f::FiniteGP, y::AbstractVecOrMat{<:Real})

...
"""
logpdf(f::FiniteGP, y::AbstractVectorMat{<:Real})
```
I would assume that this might work but I'm not sure.

theogf:
> Maybe therefore we might even want to add a more efficient implementation of loglikelihood(f::FiniteGP, Y::AbstractMatrix{<:Real}) that does not fall back to sum(logpdf(f, Y)).
@devmotion This would happen here for multi-output problems if I am not mistaking. I don't think the number of outputs will be so large that it's worth worrying about the allocation size of `logpdf(f, Y)` (which will be equal to the number of outputs)

devmotion:
The length of the vector `logpdf(f, Y)` is `size(Y, 2)`, so it does not depend on the GP but rather on the number of observations that you want to compute the likelihood/logpdfs for.

theogf:
> The length of the vector `logpdf(f, Y)` is `size(Y, 2)`, so it does not depend on the GP but rather on the number of observations that you want to compute the likelihood/logpdfs for.

Yes but in a GP you always have a fixed size number of observations (of the same size of your GP).

devmotion:
Ah yeah I guess I should not have used the term "observations". `size(Y, 2)` is the number of sets of observations, and this one is not determined by the GP. `size(Y, 1)` is fixed for each `FiniteGP` instance.

willtebbutt:
> Yes but in a GP you always have a fixed number of observations (of the same size of your GP).

There are scenarios in which you have multiple observations. A classic one would be some kind of Monte Carlo VI if your Monte Carlo approximation utilises multiple samples.

willtebbutt:
> Turing 0.14 uses loglikelihood in multiple places (e.g., https://github.com/TuringLang/DynamicPPL.jl/blob/7afd70ccab7bf76682c8b9ddb17ef4954fe56d06/src/context_implementations.jl#L449) instead of sum(logpdf(d, x)) since in particular for multi- and matrixvariate distributions the former allows for more efficient implementations that do not have to allocate (as much) (see, e.g., https://github.com/TuringLang/DistributionsAD.jl/blob/f754a19f2f8c005898674417b3cc71347526eb31/src/multivariate.jl#L146).

I see. Okay, I can definitely get on board with moving from `logpdf` to `_logpdf` then.

theogf:
Ok I think we are all on the same page :). What I meant was: In which situation do we have so many "sets of observations" (e.g. `K` sets) that allocating an array of size `K` is problematic :)

src/abstract_gp/finite_gp.jl
```diff
@@ -189,33 +189,15 @@ Random.rand(f::FiniteGP, N::Int) = rand(Random.GLOBAL_RNG, f, N)
...
-function Distributions.logpdf(f::FiniteGP, y::AbstractVector{<:Real})
-    return first(logpdf(f, reshape(y, :, 1)))
```
willtebbutt, line 216: I agree with renaming the function from `logpdf` to `_logpdf`, but why change the implementation from using the batched `logpdf` implementation?

theogf: I just thought it was an overkill to pass by the batched version?

willtebbutt: To my mind it's just good code-reuse and guarantees that the two methods remain aligned.

theogf: Sure I will change it back

src/abstract_gp/finite_gp.jl
```diff
@@ -189,6 +189,32 @@ Random.rand(f::FiniteGP, N::Int) = rand(Random.GLOBAL_RNG, f, N)
...
+julia> logpdf(f(x), y) isa Real
+true
+
+julia> Y = rand(f(x), 3);
+
+julia> logpdf(f(x), Y) isa AbstractVector{<:Real}
+true
+```
+"""
+logpdf(f::FiniteGP, y::AbstractVectorMat{<:Real})
```
devmotion, line 216:
```suggestion
logpdf(f::FiniteGP, y::AbstractVecOrMat{<:Real})
```

theogf: The dangerous ways of copy/paste :laughing:

src/abstract_gp/finite_gp.jl
```diff
@@ -189,6 +189,7 @@ Random.rand(f::FiniteGP, N::Int) = rand(Random.GLOBAL_RNG, f, N)
 Random.rand(rng::AbstractRNG, f::FiniteGP) = vec(rand(rng, f, 1))
 Random.rand(f::FiniteGP) = vec(rand(f, 1))
 
+
```
willtebbutt, line 192:
```suggestion
```

src/abstract_gp/finite_gp.jl
```diff
@@ -212,10 +213,14 @@ julia> logpdf(f(x), Y) isa AbstractVector{<:Real}
...
 ```
 """
-function Distributions.logpdf(f::FiniteGP, y::AbstractVector{<:Real})
+logpdf(f::FiniteGP, y::AbstractVecOrMat{<:Real})
+
+function Distributions._logpdf(f::FiniteGP, y::AbstractVector{<:Real})
     return first(logpdf(f, reshape(y, :, 1)))
 end
 
+Distributions.loglikelihood(f::FiniteGP, Y::AbstractMatrix{<:Real}) = sum(logpdf(f, Y))
```
willtebbutt, line 222: Is there a good way to test this that isn't just a copy + paste of the code? Would be good to ensure that this doesn't accidentally disappear in a future PR.

theogf: Oh you're absolutely right. Actually now that you are pointing it out I realize this does not even solve the initial problem cause `loglikelihood(f, y::AbstractVector)` was called.

willtebbutt: Yeah. It would probably make sense to add a minimal Turing-related integration test in this PR.

theogf: I was leaving this to #53

willtebbutt: Hm okay. If you've not got time to do it now then I think it's fine to leave it and get this merged. If you've got the time to do it though, that would be great. My concern is that it otherwise won't get done.

devmotion: The default implementation in Distributions should cover `loglikelihood(f, y::AbstractVector)`. The motivation for including this line is that the default implementation for it calls `_logpdf` for each set of observations iteratively, which leads to many unnecessary computations of the Cholesky decomposition.

theogf: Ok so I added some tests to check that `loglikelihood` does the right thing in both cases. And I also added an additional test on the Turing model described in #49. This adds a dependency on Turing but only in the tests

src/abstract_gp/finite_gp.jl
```diff
@@ -212,10 +212,14 @@ julia> logpdf(f(x), Y) isa AbstractVector{<:Real}
 true
 ```
 """
-function Distributions.logpdf(f::FiniteGP, y::AbstractVector{<:Real})
+logpdf(f::FiniteGP, y::AbstractVecOrMat{<:Real})
```
devmotion, line 215: Did you make sure that it works (i.e., that the doc string shows up)? Just wondering if it should be qualified (i.e. `Distributions.logpdf`) or not.

test/turing.jl
```diff
@@ -0,0 +1,25 @@
...
+        # Realized covariance function
+        kernel = α * transform(SqExponentialKernel(), 1/ρ)
+        f = GP(kernel)
+
+        # Sampling Distribution.
+        y ~ f(X, σ²)
+    end
+    # Test for matrices
+    m = GPRegression(y, RowVecs(X))
+    @test_nowarn sample(m, HMC(0.5, 1), 5)
```
devmotion, line 21: Wasn't this erroring before? The problem with `@test_nowarn` is that AdvancedHMC is usually quite verbose and prints many warning messages about non-finite gradients etc. So I don't think we should use it here. Maybe rather just use
```suggestion
    sample(m, HMC(0.5, 1), 5)
```
(since tests would fail if it errors) or maybe check something like
```suggestion
    @test length(sample(m, HMC(0.5, 1), 5)) == 5
```

test/turing.jl
```diff
@@ -0,0 +1,25 @@
+@testset "turing compat" begin
+    k = SqExponentialKernel()
+    y = randn(3)
+    X = randn(3, 1)
+    x = [rand(1) for _ in 1:3]
+    @model GPRegression(y, X) = begin
```
devmotion, line 6: Just a note: the `@model` syntax just expects a regular Julia function (and returns it with transformed function body), so you could equally well write
```suggestion
    @model function GPRegression(y, X)
```
The tutorials and documentation (should) use this form now as well IIRC.

test/turing.jl
```diff
@@ -0,0 +1,25 @@
...
+
+        # Sampling Distribution.
+        y ~ f(X, σ²)
+    end
+    # Test for matrices
+    m = GPRegression(y, RowVecs(X))
+    @test_nowarn sample(m, HMC(0.5, 1), 5)
+    # Test for vectors of vector
+    m = GPRegression(y, x)
+    @test_nowarn sample(m, HMC(0.5, 1), 5)
```
devmotion, line 24: Same here, maybe don't use `@test_nowarn`.

src/abstract_gp/finite_gp.jl
theogf: Yes it works nicely without the module prefix! Is there a way to test this automatically?

devmotion: One can retrieve docstrings with the `@doc` macro, see e.g. https://github.com/TuringLang/DynamicPPL.jl/blob/7afd70ccab7bf76682c8b9ddb17ef4954fe56d06/test/compiler.jl#L624-L632.

theogf: Ok the problem is that it returns the docs for all `logpdf` functions, including the ones from `Distributions.jl`, this makes it pretty much impossible to test...

devmotion: Can't you specify the arguments as well? I've never tried it though.

theogf: I tried `logpdf(::FiniteGP, ::AbstractVector)` but this only gives me the general doc from Distributions.jl

willtebbutt: `Distributions.logpder(::FiniteGP, ::AbstractVector)`?

devmotion (approved):

devmotion (changes requested):
Just found
```julia
help?> Docs.getdoc
  getdoc(obj)
  getdoc(obj, sig)

  Return a custom docstring object associated with the object obj and, optionally, the tuple type signature sig. See MultiDoc docs for an explanation of the possible
  values of sig.

  The returned object can either be a markdown object generated by Markdown.parse or some other custom type used to display non-markdown formatted documentation.

  A return value of nothing can be used to signify to the docsystem that no documentation was found for obj, in which case the docsystem will fall back to searching for
  the Binding associated with obj instead.
```
Can you add a test for the docstring using `Docs.getdoc`?

theogf:
I am not sure how to use `Docs.getdoc` : 
```julia
julia> Docs.getdoc(logpdf)
# nothing
julia> Docs.getdoc(Distributions.logpdf)
# nothing
julia> Docs.getdoc(logpdf, (AbstractGPs.FiniteGP, AbstractVector))
# nothing 
etc...
```

devmotion:
Hmm strange.... `Docs.doc` seems to work though, apparently it performs a dynamic lookup in case `Docs.getdoc` returns `nothing`. Maybe `Docs.getdoc` should just be used for types and callables structs?
```julia
julia> Docs.doc(logpdf, Tuple{AbstractGPs.FiniteGP,Vector{Float64}})
  logpdf(d::MultivariateDistribution, x::AbstractArray)

  Return the logarithm of probability density evaluated at x.

    •    If x is a vector, it returns the result as a scalar.

    •    If x is a matrix with n columns, it returns a vector r of length n, where r[i] corresponds to x[:,i].

  logpdf!(r, d, x) will write the results to a pre-allocated array r.

  logpdf(f::FiniteGP, y::AbstractVecOrMat{<:Real})

  The logpdf of y under f if is y isa AbstractVector. logpdf of each column of y if y isa Matrix.

  julia> f = GP(Matern32Kernel());

  julia> x = randn(11);

  julia> y = rand(f(x));

  julia> logpdf(f(x), y) isa Real
  true

  julia> Y = rand(f(x), 3);

  julia> logpdf(f(x), Y) isa AbstractVector{<:Real}
  true
```
That seems OK I guess. You can't get rid of the docstring for `logpdf(d::MultivariateDistributon, x::AbstractArray)` since it matches as well. Probably easiest to just check if the docs contain the string `logpdf(f::FiniteGP, y::AbstractVecOrMat{<:Real})`:
```julia
julia> contains(string(Docs.doc(logpdf, Tuple{AbstractGPs.FiniteGP,Vector{Float64}})), "logpdf(f::FiniteGP, y::AbstractVecOrMat{<:Real})")
true
```

devmotion (approved):
LGTM.

PR #58, opened 2020-09-11
Replace `contains` with `occursin`

devmotion:
`contains` requires Julia >= 1.5.


willtebbutt (approved):

willtebbutt:
Should probably have a patch version bump?

devmotion:
Even though it just fixes a test error?

willtebbutt:
Hmm yeah, maybe not. I don't have strong feelings either way.

devmotion:
Hmm but yeah on a second thought, since users might actually want to run the tests a patch version bump might be appropriate.

theogf (approved):

devmotion:
Seems we should add `print` statements to avoid time outs on MacOS?

PR #64, opened 2020-10-03
Correct the train/test data use in the last block of the ESS example

### LINKED ISSUE #63: Small error in the ESS example: Incorrect use of train/test data in the last block
This is really just a small error, but it might could actually be confusing to readers.

In the last block of code and plot in the [example on Approximate Inference with ESS](https://juliagaussianprocesses.github.io/AbstractGPs.jl/dev/examples/EllipticalSliceSampling/), there are two mistakes:
1. both scatterplots are called on `x_train` and `y_train`, so that the final plot does not actually show the test data, and
2. the posterior is actually computed with the whole dataset (`x`, `y`), which I assume should be (`x_train`, `y_train`).

I will open a very small pull request with the corrections to resolve this issue, so I guess this should be resolved very quickly.

nathanaelbosch:
Resolves #63. Please let me know if there are other additional things that should be adjusted.

### DIFF
--- a/examples/EllipticalSliceSampling.jl
+++ b/examples/EllipticalSliceSampling.jl
@@ -116,19 +116,17 @@ mean(logp(param; x=x_test, y=y_test) for param in samples)
 
 
 plt = scatter(x_train, y_train; label="Train data")
-scatter!(plt, x_train, y_train; label="Test data")
+scatter!(plt, x_test, y_test; label="Test data")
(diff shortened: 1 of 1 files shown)


sharanry (approved):
Seems fine to me. Thanks for the PR!

@willtebbutt any thoughts?

willtebbutt (approved):
Also LGTM

PR #71, opened 2020-10-22
Allow for KernelFunctions bump

willtebbutt:
Something weird happened with the accuracy of one of our tests. The numbers are quite large to I also set the rtol, seems to fix the problem.


sharanry (approved):
LGTM!

PR #74, opened 2020-11-06
Fix issues with the MeanFunctions

theogf:
This PR fixes the issue described in #14.
Similarly to what have been done in KernelFunctions.jl it replaces `Base.map` by `_map` to avoid having Zygote rules overtaking the adjoints.


willtebbutt (approved):
LGTM. Just needs a patch-version bump.

PR #90, opened 2021-01-06
Add missing whitespace

devmotion:


theogf (approved):

devmotion:
Somehow the Manifest file seems to be corrupted. I wonder why we commit `docs/Manifest.toml` at all? I guess the idea was that the docs setup is completely reproducible - but maybe it does not matter much if we keep the compat bounds updated?

devmotion:
Maybe the problem was caused by some incompatible CompatHelper PRs that both updated the Manifest file in different ways.

willtebbutt:
> Somehow the Manifest file seems to be corrupted. I wonder why we commit docs/Manifest.toml at all? I guess the idea was that the docs setup is completely reproducible - but maybe it does not matter much if we keep the compat bounds updated?

Is this not so that when the docs are built during CI we've got the Manifest pointing towards the correct version of `AbstractGPs`? Don't you get the `path = ".."` field in the Manifest entry for `AbstractGPs` or something like that?

devmotion:
It does not matter if we commit a Manifest file or not, the path of AbstractGPs will always be correct due to https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/blob/abdb04afe4cbac506c94fc44fb746eac4452d63b/.github/workflows/docs.yml#L18.

willtebbutt:
> It does not matter if we commit a Manifest file or not, the path of AbstractGPs will always be correct due to 

Oh, nice! I can't see any harm in dropping the manifest then 🤷

devmotion:
Ah OK, now I just regenerated the Manifest file :grimacing:

devmotion:
At least I noticed some typos in Pkg.jl due to the problem (https://github.com/JuliaLang/Pkg.jl/pull/2316) :smile:

devmotion:
I kept the Manifest file for now since, as mentioned above, it will yield exactly the same package environment if users want to build the documentation. I think maybe this can be beneficial for the examples which depend on a bunch of dependencies. We can still remove it in a separate PR if needed/desired.

Issue #91, opened 2021-01-06
AdvancedHMC example broken


devmotion:
For some time the AdvancedHMC example has been broken, see e.g. https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/runs/1639170780?check_suite_focus=true#step:5:56:
```julia
┌ Warning: failed to run `@example` block in src/examples/HMC.md:149-161
│ ```@example HMC
│ samples, stats = sample(
│     hamiltonian,
│     proposal,
│     initial_params,
│     n_samples,
│     adaptor,
│     n_adapts;
│     progress=false
│ )
│ samples_mat = reduce(hcat, samples)';
│ nothing #hide
│ ```
│   c.value = PosDefException: matrix is not positive definite; Cholesky factorization failed.
└ @ Documenter.Expanders ~/.julia/packages/Documenter/6vUwN/src/Expanders.jl:563
```

devmotion:
It's unstable since the implementation of `logp` is incorrect.

willtebbutt:
> It's unstable since the implementation of logp is incorrect.

Is this something in this particular example, or an issue with `AbstractGPs` itself?

devmotion:
No, just the example. I'll open a PR, I spotted some inconsistencies as well.

PR #93, opened 2021-01-06
Update examples

devmotion:
This PR updates the examples.

It contains the following changes:
- Fixes https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/issues/91 by correcting the log density - it did not take into account the prior
- Consistently uses a Gaussian noise with standard deviation 0.1 (same as GPFlow original)
- Merges the ESS and HMC example which contained identical and incorrectly copied parts
- Adds DynamicHMC example
- Uses `softplus` transformation (same as GPFlow original)

I tried to improve and add some more explanations but I am still not completely satisfied. In my opinion, it is also unsatisfying that the VI and MCMC example contain many duplicate parts, and even the different method sections in the MCMC example are very similar.


devmotion:
Maybe it would be best to merge the MCMC and VI example as well and just have one "Regression (1D)" example with different approaches and packages.

willtebbutt:
I could on board with both of your above points. Are you up for tackling the merging examples one inside this PR, or would you prefer to do it separately?

devmotion:
I'll merge them in this PR since one only has to copy the VI section to the other file.

theogf:
As an additional comment I saw multiple people getting confused with data having more than one feature, an example based on matrices/vector of vectors would also be great!

willtebbutt:
> As an additional comment I saw multiple people getting confused with data having more than one feature, an example based on matrices/vector of vectors would also be great!
A simple 2D example for example

This is a good point. I've opened [this issue](https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/issues/95) where I think that we should move the discussion to.

theogf:
> Is there any easy way to view the new documentation based on these examples, or should I pull and build locally?

I think you need to pull locally but I managed to create a PR preview for website for Franklin, maybe I can manage the same thing for documentation, but that's for another PR

devmotion:
It would be possible to set up preview builds of the documentation but currently you have to build it locally.

devmotion:
I opened a PR: https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/pull/96 We could wait for it, then it is probably easier to review the changes here.

willtebbutt:
> I opened a PR: #96 We could wait for it, then it is probably easier to review the changes here.

Unless you are desperate to get this in quickly, I'm happy to wait.

devmotion:
You can view the preview here: https://juliagaussianprocesses.github.io/AbstractGPs.jl/previews/PR93/

devmotion:
The other test error is unrelated, apparently the registry couldn't be updated: https://pipelines.actions.githubusercontent.com/ySlubJLjvdr8TQDQW4YMxXe3AcvOfty8lKuun5Y5P9Amwe0EiP/_apis/pipelines/1/runs/6350/signedlogcontent/3?urlExpires=2021-01-07T12%3A51%3A44.1666175Z&urlSigningMethod=HMACV1&urlSignature=c%2BFUN9SJCpe9Vwpsx8AzkdJTK5TAjihtdVR6eeI1dCU%3D

willtebbutt (approved):
I'm happy for this to be merged whenever you're happy @devmotion . I think the language / explanation is basically fine -- we can always iterate again later if needs be.

devmotion:
Yes, it can definitely be improved further, I was just too tired to optimize it any further at some point.

PR #102, opened 2021-01-31
Add support for FillArrays 0.11

devmotion:
Replaces https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/pull/94.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "AbstractGPs"
 uuid = "99985d1d-32ba-4be9-9821-2ec096f28918"
 authors = ["JuliaGaussianProcesses Team"]
-version = "0.2.17"
+version = "0.2.18"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
@@ -17,7 +17,7 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 [compat]
 ChainRulesCore = "0.9"
 Distributions = "0.19, 0.20, 0.21, 0.22, 0.23, 0.24"
-FillArrays = "0.7, 0.8, 0.9, 0.10"
+FillArrays = "0.7, 0.8, 0.9, 0.10, 0.11"
 KernelFunctions = "0.4, 0.5, 0.6, 0.7, 0.8"
 RecipesBase = "1"
 Reexport = "0.2"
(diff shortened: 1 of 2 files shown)


devmotion:
CI tests fail due to deprecation warnings in the doctests that are caused by the recent change from `kerneldiagmatrix` to `kernelmatrix_diag` in KernelFunctions 0.8.23. The doctests in the documentation pipeline do not fail since, apart from testing, julia does not show deprecation warnings.

Maybe we should remove the doctests from `runtests.jl`? It would be unfortunate to have to drop support for all releases of KernelFunctions apart from 0.8.23 just to update the definition of `cov_diag` (it seems `kerneldiagmatrix` is used only there).

willtebbutt:
Eurgh, that is annoying. Would the fix not be to fix the deprecations?

devmotion:
Yes, of course, one could update `cov_diag`. But then we would have to drop support for all previous versions of KernelFunctions.

> It would be unfortunate to have to drop support for all releases of KernelFunctions apart from 0.8.23 just to update the definition of cov_diag (it seems kerneldiagmatrix is used only there).

willtebbutt:
True, but it seems to me that we're not able to utilise our doctests meaningfully in our CI while we've still got deprecation warnings -- am I missing something?

devmotion:
No, that's correct - either we fix the deprecation warnings or remove doctests from `runtests.jl` (then they would be tested only in the documentation action).

willtebbutt:
But it doesn't look like they're correctly causing the documentation action to fail when they fail, meaning that we'll probably not notice if the docstests are failing. Have I understood correctly?

devmotion:
No, we will notice when the doctests fail. The only difference is that we run them with deprecation warnings being disabled (this is the default when we just call `julia -e ...`).

willtebbutt:
Ohhhhhh I see. Could we just do that for the regular CI then, rather than move the doctests out of it?

devmotion:
It seems (maybe) one can disable the deprecation warnings of the test action: https://github.com/julia-actions/julia-runtest/blob/b32c73084a913834335f8ff0ce6f6cad610a6fcf/action.yml#L18 AFAIK the tests will still fail locally though when e.g. run with `] test` since in this case deprecation warnings are enabled.

willtebbutt:
Hmmm this is a bit frustrating. Seems like there are upsides and downsides either way 😬 @theogf what are your thoughts?

theogf:
To be honest I did not get how changing to `kernelmatrix_diag` was not a breaking change and needed a version bump :) But that's irrelevant since this could happen from other packages as well.

If we notice failing in the docs that seems sufficient to me. They are **doc**tests after all :D

devmotion:
I don't think there's a problem with the version numbers of KernelFunctions here. Adding deprecations is non-breaking, only their removal is breaking.

So you would remove the doctests from `runtests.jl` and only test them (without deprecation warnings) when building the documentation @theogf ?

PR #106, opened 2021-02-17
Adds an Exact Gaussian Process Inference example to the file.

4aHxKzD:
Previously the file only had examples of approximate inference. This commit adds an example for exact inference. The example was added to the file's tail end but maybe it makes more sense to put it before the approximate inference ones.


examples/regression_1d.jl
```diff
@@ -432,3 +432,86 @@ scatter(
...
+# of the Matern kernel and different pseudo-points. We ensure that the
+# kernel parameters are positive with the softplus function ```math
+# f(x) = \log (1 + \exp x).```
+
+struct NegativeGPLoglikelihood{X,Y}
+    x::X
+    y::Y
+end
+
+function (ℓ::NegativeGPLoglikelihood)(params)
```
willtebbutt, line 457: Could we maybe just use a closure here instead of defining a new type?

willtebbutt (commented):
Thanks for contributing this! Just a quick style comment for now.

examples/regression_1d.jl
4aHxKzD: Probably, I don't have a lot of CS background (just learned about "functors"/"callable" objects while preparing this pull request) could you show me what you mean?

willtebbutt: Of course :) I was thinking something like
```julia
function negloglik_train(params)
    kernel = ScaledKernel(
        transform(
            Matern52Kernel(),
            ScaleTransform(softplus(params[1]))
        ),
        softplus(params[2]),
    )
    f = GP(kernel)
    fx = f(x_train, 0.1)
    return -logpdf(fx, y_train)
end
```
It "captures" `x_train` and `y_train` from the surrounding scope and uses them inside this function. This kind of function is known as a "closure", and it "closes over" `x_train` and `y_train`.

Under the hood, Julia's compiler more or less implements what you implemented manually with `NegativeGPLoglikelihood`, it's just doing it with a closure requires a bit less code

4aHxKzD: Thanks! I quick question of you don't mind: While this would perfectly work for this example file, doesn't it assume that `x_train` and `y_train` are global variables? I'm always reading around that we should avoid such thing.

willtebbutt: Indeed it is in general -- partly for performance reasons, partly for encapsulation reasons (globals can change without warning, which can make finding bugs in code which relies on them quite tricky). For this example code I'm not personally too worried though.

4aHxKzD: Ok, I can change it if you prefer that style.

willtebbutt: Yeah. I think it makes sense for the sake of conciseness.

4aHxKzD: Just one more thing, I initially went with this approach because the original examples use it a lot. Won't having different code styles in the same file of examples confuse novice users? On the flip side, callable objects is not a basic concept, I had spend several minutes understanding that before reading the examples, which are what really is important here.

willtebbutt: Hmmm yes, you're right -- I had forgotten that we're doing that. Let's stick with the way you're currently doing it actually, but would you mind opening an issue about this? It's helpful to know that you found it distracting from the core ideas, would be good to improve the examples :)

4aHxKzD:
Following #139 I updated this example and removed the callable struct.

4aHxKzD:
@willtebbutt can you review my code and accept the PR if everything is ok?

examples/regression_1d.jl
```diff
@@ -442,3 +442,78 @@ scatter(
...
+
+logpdf(p_fx(x_test), y_test)
+
+# We define a function which returns the negative log marginal
+# likelihood for different variance and inverse lengthscale parameters
+# of the Matern kernel and different pseudo-points. We ensure that the
+# kernel parameters are positive with the softplus function ```math
+# f(x) = \log (1 + \exp x).```
+
+function objective_function(x, y)
```
devmotion, line 462: Maybe one should rename this function to avoid confusion with the function of the same name above which is redefined here?

4aHxKzD: Sure!

4aHxKzD: Loss function?

examples/regression_1d.jl
```diff
@@ -442,3 +442,78 @@ scatter(
...
+# and obtain the following optimal parameters:
+
+θ0 = randn(2)
+opt = Optim.optimize(loss_function(x_train, y_train), θ0, LBFGS())
+
+#-
+
+opt.minimizer
+
+# The optimized value of the inverse lengthscale is
```
devmotion, line 486:
```suggestion
# The optimized value of the variance is
```

examples/regression_1d.jl
```diff
@@ -442,3 +442,78 @@ scatter(
...
+
+softplus(opt.minimizer[1])
+
+# and of the variance is
```
devmotion, line 490:
```suggestion
# and of the inverse lengthscale is
```

devmotion (commented):
In general, I wonder if one could merge this section with the VI section. It seems large parts are (almost) identical, so I assume it would be good to point out the differences more clearly.

willtebbutt:
Hmm I think I agree that @devmotion that, since the code is almost identical to the VI code, it would be good to factor the common code out to highlight the differences. I'm not going to block merging though if you feel you'd rather just get this in.

4aHxKzD:
> In general, I wonder if one could merge this section with the VI section. It seems large parts are (almost) identical, so I assume it would be good to point out the differences more clearly.

We could find a way of merging the sections. How would you call such section? 

Personally I'm not against it but not particularly in favor of it. While from a programming point of view it seems like there is a lot of repeating code, since this is an example/tutorial and not documentation I don't view it as an issue. From a conceptual point of view I think that we are separating different ways of performing inference and the user is going to click on the page's index to the one method he/she wants to use at a specific time.

I would even go as far as putting the Exact Inference at the top of the file, before the other methods, since conceptually is the first one we learn when we start learning about Gaussian Processes.

This is also how it is done in Stheno's getting started page https://juliagaussianprocesses.github.io/Stheno.jl/dev/getting_started/.

willtebbutt:
I see your point. Sounds good to me!

devmotion:
I guess this is complete?

4aHxKzD:
I'm happy enough with it.

examples/regression_1d.jl
```diff
@@ -442,3 +442,78 @@ scatter(
 scatter!(x_test, y_test; label="Test Data")
 plot!(0:0.001:1, ap; label=false)
 vline!(logistic.(opt.minimizer[3:end]); label="Pseudo-points")
+
+# ## Exact Gaussian Process Inference
+#
+# Here we use Type-II MLE to train the hyperparameters of the Gaussian process.
+# This mean that our loss function is the negative log marginal likelihood.
```
devmotion, line 449:
```suggestion
# This means that our loss function is the negative log marginal likelihood.
```

examples/regression_1d.jl
```diff
@@ -442,3 +442,78 @@ scatter(
...
+# We re-calculate the log-likelihood of the test dataset with the
+# default kernel parameters of value 1 for the sake of comparison.
+
+logpdf(p_fx(x_test), y_test)
+
+# We define a function which returns the negative log marginal
+# likelihood for different variance and inverse lengthscale parameters
+# of the Matern kernel and different pseudo-points. We ensure that the
+# kernel parameters are positive with the softplus function ```math
+# f(x) = \log (1 + \exp x).```
```
devmotion, line 460:
```suggestion
# kernel parameters are positive with the softplus function
# ``f(x) = \log (1 + \exp x)``.
```

devmotion:
Any additional comments or suggestions, @willtebbutt @theogf?

theogf:
Looks good to me! LGTM

theogf (approved):

willtebbutt (approved):
Also LGTM

PR #115, opened 2021-03-29
Remove `Symmetric(::Diagonal)` type piracy

devmotion:
As the name says, this removes some pretty severe type piracy. Instead I switched to `_symmetric` wherever it seems possible to operate with diagonal matrices, and moved the helper functions `_symmetric` and `_cholesky` and their tests to the `src/util` folder.


willtebbutt (approved):
Thanks for doing this -- infinitely better than before.

PR #116, opened 2021-03-29
Implement `cov_diag` and `mean_and_cov_diag` for `FiniteGP`s

devmotion:
This PR
- implements `cov_diag` and `mean_and_cov_diag` for `FiniteGPs` and uses it to compute `marginals`
- adds default implementations of ~~`cov_diag`~~ (see below), `mean_and_cov`, and `mean_and_cov_diag`
- exports `mean_and_cov_diag`

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "AbstractGPs"
 uuid = "99985d1d-32ba-4be9-9821-2ec096f28918"
 authors = ["JuliaGaussianProcesses Team"]
-version = "0.2.20"
+version = "0.2.21"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "AbstractGPs"
 uuid = "99985d1d-32ba-4be9-9821-2ec096f28918"
 authors = ["JuliaGaussianProcesses Team"]
-version = "0.2.19"
+version = "0.2.20"
 
 [deps]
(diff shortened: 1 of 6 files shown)


willtebbutt:
I like this.

Before properly reviewing, the only thing that concerns me in this PR is the addition of the default for `cov_diag`. In my experience, I would much rather `cov_diag` fails if I've not got an efficient way to do it so as to avoid hitting a fallback with asymptotically worse complexity than what I would usually expect to get.

Is there a particular reason that you want this fallback implementation?

devmotion:
No, there's no particular reason apart from the fact that it could be added and it would reduce the number of methods one has to implement for a subtype of `AbstractGP`. However, I was also debating if it would be better to add default implementations only for `mean_and_cov` and `mean_and_cov_diag` such that it obvious that it uses the slow fallback. On the other hand, the default implementation of `mean_and_cov` and `mean_and_cov_diag` (and the existing implementation for `GP`) also use the slow fallback and it's not obvious to users that this is the case (well, at least not for `GP`).

willtebbutt:
> On the other hand, the default implementation of mean_and_cov and mean_and_cov_diag (and the existing implementation for GP) also use the slow fallback and it's not obvious to users that this is the case (well, at least not for GP).

Right, yes. For me, the distinction is that the default for `mean_and_cov` and `mean_and_cov_diag` is (roughly speaking) at worst 2x slower than a fast method. Conversely, the default implementation for `cov_diag` is potentially orders of magnitude slower.

devmotion:
True, I'll remove it and if needed it could be added at some point.

willtebbutt (approved):
LGTM

Issue #113, opened 2021-03-28
Specialised Plotting Inputs for FiniteGP


willtebbutt:
Our plotting functionality currently assumes that when plotting a `FiniteGP` `f(x)`, we wish to visualise it at the inputs `x`. While this assumption generally holds, there's currently no escape hatch (beyond circumventing the plotting functionality entirely!) if this doesn't hold.

My particular use case is that I have a `FiniteGP` `f(x)` whose dimensions I wish to plot on the x-axis at some other set of points `z`.

I'm not super familiar with plotting recipes, @theogf or @devmotion , do you know how I might we might extend our existing functionality to make this work?

devmotion:
I could try to extend the plotting recipes but unfortunately it is not completely clear to me what additional functionality you would like to have.

Currently, we just plot the marginals of a `FiniteGP`, which is a projection of a GP at some fixed points `x`, versus the points `x`. Can you explain maybe with an example what you would like to plot instead?

willtebbutt:
Great. So, concrete example:
```julia
z = randn(10)
x = randn(10)
f = GP(SEKernel())
plot(z, f(x))
```
should produce something similar to
```julia
plot(z, mean(f(x)); ribbon=std(f(x)))
```
(modulo our usual formatting defaults)

devmotion:
Great, I think I know how the recipe has to be extended to account for this case. I'll prepare a PR.

theogf:
Could you explain what's your use case? I don't understand why you need this.

willtebbutt:
The situation I've got is analogous to wanting to plot a single output from a multi-output GP. The inputs `x` for such a GP will be a vector of `Tuple`s (or some abstraction based around that does actually instantiate said vector fully), the first element of which is the index of the output, the second the location in space on which your multi-output GP lives.

So the current plotting functionality won't know how to plot `f(x)`, because it doesn't know how to handle `x`, but with this addition I'll be able to say `plot(last.(x), f(x))` and get the thing that I need.

(I'm not actually doing this with multi-output GPs, I'm refactoring Stheno to use AbstractGPs properly and trying to get AbstractGPs' plotting functionality to play nicely with that)

PR #118, opened 2021-03-29
Update plot recipe and allow to specify independent `x` variable

devmotion:
Fixes https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/issues/113 without breaking existing plotting functionality.

I am not completely happy with the `AbstractGP` dispatches though. I think maybe the following breaking changes would be a bit more intuitive (even without the additional option to specify the `x` variables):
- change `plot(gp::AbstractGP, x::AbstractVector)` to `plot(x, gp)` and forward it to `plot(x, gp(x))`: the points `x` of the projection are used as the first variable in the plot
- remove `plot(gp::AbstractGP, xmin::Real, xmax::Real)`: IMO there's no clear advantage over specifying `x` directly and it seems inconsistent with the Plots ecosystem
- only add a dispatch for `plot(x::AbstractVector, gp::FiniteGP)` but not one for `plot(x::AbstractVector, gp::AbstractGP)`: IMO the use case of different points `x` and `gp.x` seems quite special and therefore it seems reasonable to keep the dispatches and plotting signatures simple and demand that users construct the `FiniteGP` object manually

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "AbstractGPs"
 uuid = "99985d1d-32ba-4be9-9821-2ec096f28918"
 authors = ["JuliaGaussianProcesses Team"]
-version = "0.2.21"
+version = "0.2.22"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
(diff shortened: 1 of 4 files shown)


willtebbutt (approved):
I agree that your proposals seem reasonable. I also agree that the range stuff doesn't add a huge amount.

In the mean time, I think this is good to go.

PR #121, opened 2021-03-30
Reorganize plotting commands

devmotion:
This PR reorganizes plotting commands as suggested in https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/pull/118:
- `plot(x::AbstractVector, gp::FiniteGP)` plots `mean(gp)` vs `x` with a band of one standard deviation `sqrt.(cov_diag(gp))`
- `plot(gp::FiniteGP)` is forwarded to `plot(gp.x, gp)`
- `plot(x::AbstractVector, gp::AbstractGP)` is forwarded to `plot(gp(x))`

The following calls are deprecated:
- `plot(gp::AbstractGP, x::AbstractVector)` (replaced with `plot(x, gp)`)
- `plot(gp::AbstractGP, xmin::Real, xmax::Real)` (replaced with `plot(range(xmin, xmax; length=1_000), gp)`)
- `plot(z::AbstractVector, gp::AbstractGP, x::AbstractVector)` (replaced with `plot(z, gp(x))`)
- `plot(z::AbstractVector, gp::AbstractGP, xmin::Real, xmax::Real)` (replaced with `plot(z, gp(range(xmin, xmax; length=1_000)))`)

### DIFF
--- a/src/util/plotting.jl
+++ b/src/util/plotting.jl
@@ -3,12 +3,13 @@
 @recipe function f(x::AbstractVector, gp::FiniteGP)
     length(x) == length(gp.x) ||
         throw(DimensionMismatch("length of `x` and `gp.x` has to be equal"))
+    scale = pop!(plotattributes, :ribbon_scale, 1.0)::Float64
+    scale > 0.0 || error("`bandwidth` keyword argument must be non-negative")
 
     # compute marginals
     μ, σ2 = mean_and_cov_diag(gp)
-    σ = map(sqrt, σ2)
 
-    ribbon := σ
+    ribbon := scale .* sqrt.(σ2)
     fillalpha --> 0.3
     linewidth --> 2
     return x, μ
--- a/src/util/plotting.jl
+++ b/src/util/plotting.jl
@@ -3,7 +3,7 @@
 @recipe function f(x::AbstractVector, gp::FiniteGP)
     length(x) == length(gp.x) ||
         throw(DimensionMismatch("length of `x` and `gp.x` has to be equal"))
-    scale = pop!(plotattributes, :ribbon_scale, 1.0)::Float64
+    scale::Float64 = pop!(plotattributes, :ribbon_scale, 1.0)
     scale > 0.0 || error("`bandwidth` keyword argument must be non-negative")
 
     # compute marginals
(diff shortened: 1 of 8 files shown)


willtebbutt:
I'm pro- these changes.

At the risk of asking for more, any chance we could have some control over the number of std devs that are plotted? I personally find 3 more helpful than 1 😂

devmotion:
Sure, what would be a good name for such a parameter? `bandwidth`?

willtebbutt:
Seems reasonable to me. Maybe `ribbon_scale`? Would relate nicely to the `ribbon` kwarg?

src/util/plotting.jl
```diff
@@ -1,24 +1,18 @@
...
-
+@recipe f(x::AbstractVector, gp::AbstractGP) = gp(x)
 @recipe f(gp::FiniteGP) = (gp.x, gp)
-@recipe function f(z::AbstractVector, gp::FiniteGP)
-    length(z) == length(gp.x) ||
-        throw(DimensionMismatch("length of `z` and `gp.x` has to be equal"))
+@recipe function f(x::AbstractVector, gp::FiniteGP)
+    length(x) == length(gp.x) ||
+        throw(DimensionMismatch("length of `x` and `gp.x` has to be equal"))
+    scale = pop!(plotattributes, :ribbon_scale, 1.0)::Float64
```
willtebbutt, line 6: Has to be `Float64`, or would `Real` suffice?

devmotion: `Real` would be sufficient (I assume) but I wanted to avoid abstract types.

willtebbutt: Is there a particular reason to avoid abstract types here?

willtebbutt: I see. I would quite like to be able to write `ribbon_scale=3` rather than `ribbon_scale=3.0`, but I'm also not overly fussed.

devmotion: I guess we could change it to
```suggestion
    scale::Float64 = pop!(plotattributes, :ribbon_scale, 1.0)
```
instead? That would perform a conversion to `Float64` first before checking the type assertion. In fact that was what I wanted to do here but I'm always confused about where to put the type assertions :smile:

theogf: The Int vs Float64 is a real pitfall I agree with @willtebbutt that it's probably better to have a Real.

willtebbutt: > That would perform a conversion to Float64 first before checking the type assertion.

I had absolutely no idea that that's how that works. I'm definitely happy for stuff to be converted to a `Float64` internally.

willtebbutt (approved):
LGTM!

theogf (approved):

Issue #122, opened 2021-03-30
 MethodError: no method matching bijector(::AbstractGPs.FiniteGP)


andreaskoher:
Hey all,

first of all, great package! I work a lot with Turing and now try to add GPs in order to estimate latent functions. That said, the simple example below breaks
```julia
using AbstractGPs, KernelFunctions
using Turing

X = 1:100
f = GP( Matern32Kernel() )
u = rand( f(X) )
λ = exp.(u)
y = rand.(Poisson.(λ))

@model function latent_gp_model(X, y)
    f  = GP(Matern32Kernel())
    u ~ f(X)
    λ = exp.(u)
    y .~ Poisson.(λ)
end
c = sample( latent_gp_model(X, y), NUTS(), 10)
```
>  MethodError: no method matching bijector(::AbstractGPs.FiniteGP)

As a workaround, I can define a struct that inherits from AbstractMvNormal and everything works fine:

```julia
struct MvNormalGP{T<:AbstractGPs.FiniteGP} <: AbstractMvNormal
    f::T
end
AbstractGPs.rand(rng::Random.AbstractRNG, f::MvNormalGP) = rand(rng, f.f)
AbstractGPs.logpdf(f::MvNormalGP, y::AbstractVector{<:Real}) = logpdf(f.f, y)

@model function latent_gp_model(X, y)
    f  = GP(Matern32Kernel())
    u ~ MvNormalGP( f(X) )
    λ = exp.(u)
    y .~ Poisson.(λ)
end
c = sample( latent_gp_model(X, y), NUTS(), 10)
```

Thats not a serious issue but, as a beginner to both Turing and AbstractGPs, it took me a while to figure out the problem. It is also somewhat inconsistent in the case of Stheno: With the recent [PR](https://github.com/JuliaGaussianProcesses/Stheno.jl/pull/136), sampling of sparse latent GPs with Stheno works out of the box but, sampling from a non-sparse GP has the same issue as above since the implementation is based on AbstractGPs.

Do you think, it would be possible to derive GPs from AbstractMvNormal instead of ContinuousMultivariateDistribution (at least in the case of Gaussian likelihoods) ?

willtebbutt:
Seconded. We should definitely be doing that. I think when I implemented this originally I was unaware that `Distributions` provided an `AbstractMvNormal` type.

andreaskoher:
Cool. Should I make a PR?

willtebbutt:
[...]

PR #123, opened 2021-03-30
Derive FiniteGP from AbstractMvNormal

andreaskoher:
This PR fixes [issue #122](https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/issues/122). 

The only change is that FiniteGP is now derived from AbstractMvNormal instead of ContinuousMultivariateDistribution. 
Now, we can now directly sample from a latent finite GP in Turing as follows:

```julia
using AbstractGPs, KernelFunctions
using Turing

X = 1:100
f = GP( Matern32Kernel() )
u = rand( f(X) )
λ = exp.(u)
y = rand.(Poisson.(λ))

@model function latent_gp_model(X, y)
    f  = GP(Matern32Kernel())
    u ~ f(X)
    λ = exp.(u)
    y .~ Poisson.(λ)
end
c = sample( latent_gp_model(X, y), NUTS(), 10)
```

It's a somewhat trivial change, so I have not added any tests. Is there anything I missed?


devmotion:
Can you maybe add the Turing example to the Turing tests?

willtebbutt:
Do we want to consider this to be breaking? I guess stuff could in principle break as a consequence of this...

devmotion:
Also related: https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/issues/124

andreaskoher:
> Can you maybe add the Turing example to the Turing tests?

ah right, will do.

test/turing.jl
```diff
@@ -1,25 +1,44 @@
...
+			u ~ f(X)
+			λ = exp.(u)
+			y .~ Poisson.(λ)
+		end
+		m = latent_gp_regression(y, RowVecs(X))
+		@test length(sample(m, NUTS(), 5)) == 5
+		# Test for vectors of vector
+		m = latent_gp_regression(y, x)
+		@test length(sample(m, NUTS(), 5)) == 5
+	end
```
devmotion, line 43: I think we should avoid tabs:
```suggestion
        X = randn(3, 1)
        x = [rand(1) for _ in 1:3]
        y = rand.(Poisson.(exp.(randn(3))))

        @model function latent_gp_regression(y, X)
            f  = GP(Matern32Kernel())
            u ~ f(X)
            λ = exp.(u)
            y .~ Poisson.(λ)
        end
        m = latent_gp_regression(y, RowVecs(X))
        @test length(sample(m, NUTS(), 5)) == 5
        # Test for vectors of vector
        m = latent_gp_regression(y, x)
        @test length(sample(m, NUTS(), 5)) == 5
    end
```

devmotion:
It seems the only thing left to do is to update the version number?

willtebbutt:
Do we want to clean out our deprecations before releasing 0.3, or leave them in until 0.4? They've not been around for a particularly large amount of time, so possibly we want to keep them around for a bit?

devmotion:
Hmm I don't know, I guess it's your call: according to JuliaHub only your packages depend on AbstractGPs (https://juliahub.com/ui/Packages/AbstractGPs/NOFPh/0.2.25?t=2) and I assume that not (too) many users work with it directly yet.

willtebbutt:
Great, lets get rid of the deprecations then -- I've already upgraded Stheno anyway 🤷

willtebbutt (approved):
LGTM.

devmotion:
Thank you for your contribution, @andreaskoher!

Issue #124, opened 2021-03-30
Rename `cov_diag` to `var`?


devmotion:
Distributions.jl uses `var` for the diagonal of the covariance matrix and, e.g., defines a fallback for `AbstractMvNormal`: https://github.com/JuliaStats/Distributions.jl/blob/35125579a77b938f8d4a8bd2be23cbf1f2ddf225/src/multivariate/mvnormal.jl#L258

So to avoid that users end up with surprisingly inefficient code we could
- define `Statistics.var(gp::FiniteGP) = cov_diag(gp)`
- or rename `cov_diag` to `var`

What would you prefer?

theogf:
Renaming `cov_diag` to `Statistics.var` sounds great to me!

willtebbutt:
If we were to do this, I'd want to rename all of the places where we refer to diagonals, such as `mean_and_cov_diag` -> `mean_and_var` to ensure that we're consistent.

devmotion:
Yes, I agree, I thought about this as well and wanted to suggest it but somehow forgot to mention it above :smile:

PR #125, opened 2021-03-30
Replace `cov_diag` and `mean_and_cov_diag` with `var` and `mean_and_var`

devmotion:
Fixes https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/issues/124.

### DIFF
--- a/src/AbstractGPs.jl
+++ b/src/AbstractGPs.jl
@@ -12,7 +12,7 @@ module AbstractGPs
 
     using KernelFunctions: ColVecs, RowVecs
 
-    export GP, mean, cov, std, cov_diag, mean_and_cov, mean_and_cov_diag, marginals,
+    export GP, mean, cov, var, std, mean_and_cov, mean_and_var, marginals,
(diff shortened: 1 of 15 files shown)


src/AbstractGPs.jl
```diff
@@ -12,7 +12,7 @@ module AbstractGPs
 
     using KernelFunctions: ColVecs, RowVecs
 
-    export GP, mean, cov, std, cov_diag, mean_and_cov, mean_and_cov_diag, marginals,
+    export GP, mean, cov, var, std, mean_and_cov, mean_and_var, marginals,
```
theogf, line 15: Even though it's deprecated shouldn't we still export the cov_diag functions? Or does @deprecated does that by itself?

theogf (approved):

src/AbstractGPs.jl
devmotion: Yep, it's exported automatically by `@deprecated`.

PR #126, opened 2021-03-31
Inheriting mean_and_cov mean_and_var from StatsBase

theogf:
I don't know if it's something we want but `mean_and_var` and `mean_and_cov` already exist in `StatsBase.jl` and I thought it would make sense to have our function inherit from them?


willtebbutt:
In principle it seems okay, but I'm not thrilled at the prospect off adding a dependency on `StatsBase`, as it has quick a few dependencies itself.

theogf:
I still don't understand why there is Statistics AND StatsBase :laughing:

devmotion:
I agree it would be good to minimize the number of dependencies. However, it seems we already depend on StatsBase indirectly via Distributions?

willtebbutt:
You're totally right. Yeah, lets do it.

willtebbutt (approved):
Just needs a patch bump. I don't _think_ that this is breaking...

devmotion (approved):

theogf:
@willtebbutt I think this PR and #125 might cause trouble in Stheno in this file : https://github.com/JuliaGaussianProcesses/Stheno.jl/blob/master/src/abstract_gp.jl

devmotion:
I don't think the changes in https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/pull/125 were breaking - you can still use the old names since they are exported and deprecated correctly (hopefully :slightly_smiling_face:). And it seems the changes in this PR would only be breaking if you implement both `StatsBase.mean_and_var` and `AbstractGPs.mean_and_var` (or the old `AbstractGPs.mean_and_cov_diag`) in different ways?

theogf:
Ah yeah I missed the import... My bad

PR #127, opened 2021-04-03
Fix plotting example in README.md

devmotion:
Just noticed that the README contains an outdated example.


willtebbutt (approved):
LGTM, sorry for taking a while to review!

PR #128, opened 2021-04-04
Add basic API docs

willtebbutt:
This is both documentation and a naming proposal. So feedback on the way in which I've chopped things up / expressed things would be very helpful.


theogf:
Could you do the same thing as in KernelFunctions.jl and allow PR docs builds so we can preview the changes

devmotion:
> Could you do the same thing as in KernelFunctions.jl and allow PR docs builds so we can preview the changes

It is already enabled IIRC.

docs/make.jl
```diff
@@ -42,11 +42,12 @@ makedocs(;
...
         "Home" => "index.md",
-        "API" => "api.md",
+        "The Main APIs" => "api.md",
+        "Concrete Features" => "concrete_features.md",
         "Examples" => joinpath.("examples", filter(x -> endswith(x, ".md"), readdir(OUTPUT))),
     ],
-    strict=true,
-    checkdocs=:exports, 
+    # strict=true,
+    # checkdocs=:exports,
```
devmotion, line 50: What is the reason for commenting out these checks?

docs/src/api.md
```diff
@@ -1,5 +1,115 @@
-# API
+# FiniteGPs and AbstractGPs
+
+There are two core types that AbstractGPs provides: `FiniteGP`, and `AbstractGP`.
```
devmotion, line 3: It seems a bit strange the the two core types are not exported, isn't it?

docs/src/api.md
```diff
@@ -1,5 +1,115 @@
...
+
+
+## FiniteGP
+
+AbstractGPs.jl provides the `FiniteGP` type. It represents the multivariate Gaussian induced by "indexing" into a GP `f` at a collection of points `x`, and adding independent zero-mean noise with variance `Σ`:
+```julia
+fx = f(x, Σ)
+
+# The above is equivalent to:
+fx = AbstractGPs.FiniteGP(f, x, Σ)
```
devmotion, line 13: Maybe only one construction should be allowed and/or recommended?

docs/src/api.md
```diff
@@ -1,5 +1,115 @@
...
+```
+
+The `FiniteGP` has two API levels.
+The first should be supported by all `FiniteGPs`, while the second will only be supported by a subset.
```
devmotion, line 17: Matter of taste but I prefer to use inline code only for correct Julia code:
```suggestion
The first should be supported by all `FiniteGP`s, while the second will only be supported by a subset.
```

docs/src/api.md
```diff
@@ -1,5 +1,115 @@
...
+
+
+
+### Primary Public API
+
+These are user-facing methods.
+You can expect them to be implemented whenever you encounter a `FiniteGP`.
+If you are building something on top of AbstractGPs.jl, try to implement it in terms of these functions.
```
devmotion, line 25: Maybe we should consistently write AbstractGPs.jl or AbstractGPs? I don't mind which alternative we use but it would be nice to be consistent.

docs/src/api.md
```diff
@@ -1,5 +1,115 @@
...
+#### Optional Methods
+Default implementations are provided for these, but you may wish to specialise for performance.
+```@docs
+mean_and_cov(::AbstractGPs.FiniteGP)
+```
+
+
+## Internal AbstractGPs API
+
+This functionality is not intended to be used directly be users, or those building functionality on top of `AbstractGP` -- they should interact with Primary Public API above, and the Seconary Public API if truly necessary.
```
devmotion, line 72:
```suggestion
This functionality is not intended to be used directly be users, or those building functionality on top of `AbstractGP` -- they should interact with Primary Public API above, and the Secondary Public API if truly necessary.
```

devmotion: And maybe one could insert links to the two API sections above?

docs/src/api.md
```diff
@@ -1,5 +1,115 @@
...
+
+#### Optional Methods
+Default implementations are provided for these, but you may wish to specialise for performance.
+```@docs
+cov(::AbstractGPs.AbstractGP, ::AbstractVector)
+mean_and_cov(::AbstractGPs.AbstractGP, ::AbstractVector)
+mean_and_var(::AbstractGPs.AbstractGP, ::AbstractVector)
+```
+
+Note that, while we _could_ provide a default implementation for `var` as `diag(cov(f, x))`, this is generally such an inefficient fallback, that we view find it preferable to error if it's not implemented than to ever hit a fallback.
```
devmotion, line 96:
```suggestion
Note that, while we _could_ provide a default implementation for `var(f, x)` as `diag(cov(f, x))`, this is generally such an inefficient fallback, that we find it preferable to error if it's not implemented than to ever hit a fallback.
```

docs/src/api.md
```diff
@@ -1,5 +1,115 @@
...
+
+
+
+## When to implement FiniteGP methods
+
+If you have a new subtype of `AbstractGP` and it implements the API above (i.e. you don't mind computing covariance matrices), then you'll not usually need to add new methods involving your own `FiniteGP` -- the fallback implementations will often be completely fine.
+
+If, on the other hand, you don't want to implement the Internal AbstractGPs API for e.g. performance reasons, then you'll need to implement the Primary Public API directly.
```
devmotion, line 104: Again maybe one could link to the corresponding sections?

docs/src/api.md
```diff
@@ -1,5 +1,115 @@
...
+```
+
+
+
+### Secondary Public API
+
+While the covariance matrix of any multivariate Gaussian is defined, it is not always a good idea to actually compute it.
+Fortunately, it's often the case that you're not actually interested in the covariance matrix per-se, rather the other quantities that you might use it to compute (`logpdf`, `rand`, `posterior`).
+This is similar to the well-known observation that you rarely need the inverse of a matrix, you just need to compute the inverse multiplied by something, so it's considered good practice to avoid ever explicitly computing the inverse of a matrix so as to avoid the numerical issues associated with it.
+This is important, for example, as `TemporalGPs.jl` is able to implement all of the Primary Public API in linear time in the dimension of the `FiniteGP`, as it never needs to evaluate the covariance matrix.
```
devmotion, line 52: Maybe add a link to TemporalGPs? Another slight inconsistency here is that you write `TemporalGPs.jl` (as inline code) but used AbstractGPs and AbstractGPs.jl before (as regular text).

docs/src/api.md
```diff
@@ -1,5 +1,115 @@
...
+Note that, while we _could_ provide a default implementation for `var` as `diag(cov(f, x))`, this is generally such an inefficient fallback, that we view find it preferable to error if it's not implemented than to ever hit a fallback.
+
+
+
+## When to implement FiniteGP methods
+
+If you have a new subtype of `AbstractGP` and it implements the API above (i.e. you don't mind computing covariance matrices), then you'll not usually need to add new methods involving your own `FiniteGP` -- the fallback implementations will often be completely fine.
+
+If, on the other hand, you don't want to implement the Internal AbstractGPs API for e.g. performance reasons, then you'll need to implement the Primary Public API directly.
+This is the case in `TemporalGPs.jl` -- the covariance matrix is never actually needed, so we neglect to provide any implementations involving it, instead implementing specialised methods for the Primary Public API.
```
devmotion, line 105: Maybe some links to TemporalGPs and the API? And maybe don't use inline code for TemporalGPs.jl?

docs/src/concrete_features.md
```diff
@@ -0,0 +1,98 @@
+## Features
+
+### Setup
+```julia
```
devmotion, line 4: Could we use `@example` blocks here to make sure that the code works?

devmotion: Ah wait, this is only copied from the current landing page, right? Then maybe we don't have to address it in this PR but it should definitely be improved (not only headings, some explanations etc.).

devmotion:
I like the explanations of the API (it is very much needed IMO) and it seems reasonable to me. I can't really comment on any issues/inconsistencies one might face in practice since I haven't tried to implement something on top of it.

docs/make.jl
willtebbutt: I've removed the autodocs stuff, meaning that a lot of stuff that was in the docs via autodocs is no longer present. My feeling is that it's preferable to have no docs than to just have the docstrings without any context. Happy to re-add the autodocs if we prefer having it.

docs/src/api.md
willtebbutt: Good point. My intention is that the former is preferred -- I've just documented the latter to make it clear what the former is doing. Would a comment along the lines of "This is just for reference -- please use the above syntax" be sufficient?

docs/src/concrete_features.md
willtebbutt: Indeed -- tidying up this bit of the docs is for a different PR. I just wanted to stop this from being the first thing that you see when you land in the docs.

docs/src/api.md
willtebbutt: I kind of agree. At present, AbstractGPs serves two competing purposes -- it both specifies an API, and provides a few concrete implementations of said API (e.g. `GP`). Users of said concrete implementations shouldn't ever need to interact with either `AbstractGP`s or `FiniteGP`s, just things like `GP`.

If we split this package up to separate the APIs from the implementations, it would probably make sense to export `AbstractGP` and `FiniteGP`.

Perhaps I should comment on this in the docs?

docs/src/api.md
devmotion: Yes :+1:

docs/src/api.md
devmotion: Yes, I guess a short comment similar to your explanation here would be good.

docs/make.jl
devmotion: No, I think it's completely fine to remove autodocs. I agree that docstrings without context are not helpful. However, I think at some point we should add them back *with* context - IMO usually the documentation should include docstrings of all exported functions and types.

willtebbutt: I agree entirely -- we need another couple of PRs of roughly this size to finish off documenting the basics of the API.

theogf (commented):
So on my side, I feel that the explanations are still very abstract and not especially newcomer friendly. 
- I think what's lacking is a more proper explanation of what these two types represent mathematically.
- Additionally I think the explanation on FiniteGP can be misleading. It makes it sound that it's an abstract type that custom type should inherit from. Whereas one should overload the methods for the different **parametrizations** of `FiniteGP`

docs/src/api.md
```diff
@@ -1,5 +1,116 @@
-# API
+# FiniteGPs and AbstractGPs
 
-```@autodocs
-Modules = [AbstractGPs]
-```
\ No newline at end of file
+There are two core abstract types that AbstractGPs provides: `FiniteGP`, and `AbstractGP`.
```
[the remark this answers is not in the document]
willtebbutt: Good point, oversight on my part. Will fix.

docs/src/api.md
```diff
@@ -1,5 +1,116 @@
...
+# When writing code, prefer the above syntax.
+fx = AbstractGPs.FiniteGP(f, x, Σ)
+```
+
+The `FiniteGP` has two API levels.
+The first should be supported by all `FiniteGP`s, while the second will only be supported by a subset.
+
+
+
+### Primary Public API
```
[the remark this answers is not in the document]
willtebbutt: I went with the names "Primary" and "Secondary" for lack of better terms. Perhaps "FiniteGP API" and "Extended FiniteGP API" would make things clearer? i.e. all `FiniteGP`s implement the "FiniteGP API", but only some implement the "Extended FiniteGP API", and whether or not the "Extended FiniteGP API" is implemented depends on the `AbstractGP` contained inside?

docs/src/api.md
```diff
@@ -1,5 +1,116 @@
...
+```
+
+#### Optional Methods
+Default implementations are provided for these, but you may wish to specialise for performance.
+```@docs
+mean_and_cov(::AbstractGPs.FiniteGP)
+```
+
+
+## Internal AbstractGPs API
```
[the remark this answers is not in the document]
willtebbutt: This is internal from a user's perspective, and by user I mean someone who is trying to use GPs to model something as opposed to someone who is implementing the API.

I think this is one instance where having both implementations and API in the same package is starting to cause confusion.

docs/src/api.md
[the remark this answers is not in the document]
willtebbutt: Added. Let me know if you think it's clearer now.

docs/src/api.md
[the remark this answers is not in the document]
theogf: Yeah I see your point and realised what was meant in my general review comment. This should in any case be properly explained.
In this sense I would probably even start by the `AbstractGP` interface before the `FiniteGP` one.

docs/src/api.md
```diff
@@ -1,5 +1,116 @@
...
+
```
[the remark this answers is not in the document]
willtebbutt: Would you like an explanation of what an AbstractGP is here as well, or is the one added at the top sufficient?

docs/src/api.md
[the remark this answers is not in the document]
willtebbutt: > In this sense I would probably even start by the AbstractGP interface before the FiniteGP one.

Yeah, this is where it starts to get slightly tricky, so I think we should think a bit about how to make this work.

The reason that I started with the `FiniteGP` API is because it's the thing that you can use to compute stuff. It turns out that an `AbstractGP` on its own is not especially useful, it's only when you bring in some inputs `x` that you can start to compute things. So in that sense, if you've defined a new `AbstractGP`, the thing that you really want implemented is the FiniteGP API for `FiniteGP{<:MyNewAbstractGP}`. Perhaps I should lead with that?

willtebbutt:
I've re-drafted quite a bit of this. Would be a good time to do another pass.

docs/src/api.md
```diff
@@ -1,5 +1,167 @@
...
\ No newline at end of file
+AbstractGPs provides one main abstract type, the `AbstractGP`.
+Instances of subtypes of `AbstractGP` represent Gaussian processes -- collections of jointly-Gaussian random variables, which may be infinite-dimensional.
+AbstractGPs only become useful computationally when you specify a finite-dimensional marginal distribution to work with.
+It is for this reason that we provide the `FiniteGP` type.
+Roughly speaking, this type comprises an `AbstractGP` `f` and an `AbstractVector` `x`, and represents the multivariate Gaussian distribution over `f` at `x`.
+This distribution, being finite-dimensional, is something that can be used to compute useful things.
+For example, `rand(f(x))` generates a sample from the multivariate Gaussian that is `f` at `x`, and `logpdf(f(x), y)` computes the log (marginal) probability of observing the vector `y` under this distribution.
+
+Consequently, if you create a new `AbstractGP` subtype, say you called it `MyNewGP`, the methods that you must ensure are implemented are not those directly involving `MyNewGP`, but rather
```
theogf, line 11:
```suggestion
Consequently, if you create a new subtype `MyNewGP` of `AbstractGP`, the methods that you must implement are not those directly involving `MyNewGP`, but rather those involving
```

docs/src/api.md
```diff
@@ -1,5 +1,167 @@
...
+```julia
+FiniteGP{<:MyNewGP}
+```
+We provide two ways in which to do this.
```
theogf, line 15: This is contradicting the sentence before "you must ensure are implemented"

docs/src/api.md
```diff
@@ -1,5 +1,167 @@
...
+The first is to implement methods directly on `Finite{<:MyNewGP}` -- this is detailed in the [FiniteGP APIs](@ref).
+The second is to implement some methods directly involving `MyNewGP`, and utilise default `FiniteGP` methods implemented in terms of these -- this is detailed in the [Internal AbstractGPs API](@ref).
+
+The second interface is generally easier to implement, but isn't applicable for all subtypes of `AbstractGP`.
+See [Which API should I implement?](@ref) for further discussion.
+
+Note that neither `AbstractGP` nor `FiniteGP` are exported as this package also provides user-facing concrete types which are generally what users should interact with.
+Package developers, and anyone writing code that is intended to work with any GP in the ecosystem, should import what they need.
```
theogf, line 23:
```suggestion
Package developers, and anyone writing code that is intended to work with any GP in the ecosystem, should build upon this two types.
```

docs/src/api.md
```diff
@@ -1,5 +1,167 @@
...
+
+## Intended Audience
+
+If you do not plan to implement your own `AbstractGP` subtype because e.g. you just wish to build a model involving a GP, then the useful part of the API for you is the [FiniteGP APIs](@ref).
```
theogf, line 27: "involving a GP" is extremely unclear in this context.
Since this whole page is aimed at developers, I think we should explicitely say it at the beginning and redirect more standard users to the examples.

docs/src/api.md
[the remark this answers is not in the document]
theogf: I think the one you added at the top is now sufficient. It cannot hurt to have it twice though :)

docs/src/api.md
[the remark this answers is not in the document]
theogf: I think starting with a very concrete example would be quite self-explanatory. For example showing that for `AbstractGP` one would implement `mean(f, x)` but for `FiniteGP` it's by default `mean(f.gp, f.x)` or one can overload it.

docs/src/api.md
```diff
@@ -1,5 +1,167 @@
...
+cov(::AbstractGPs.AbstractGP, ::AbstractVector)
+mean_and_cov(::AbstractGPs.AbstractGP, ::AbstractVector)
+mean_and_var(::AbstractGPs.AbstractGP, ::AbstractVector)
+```
+
+Note that, while we _could_ provide a default implementation for `var(f, x)` as `diag(cov(f, x))`, this is generally such an inefficient fallback, that we find it preferable to error if it's not implemented than to ever hit a fallback.
+
+
+
+## Which API should I implement?
```
theogf, line 136: That's a great addition!

docs/src/api.md
```diff
@@ -1,5 +1,167 @@
-# API
+# FiniteGPs and AbstractGPs
```
theogf, line 1:
```suggestion
# FiniteGP and AbstractGP
```
I think the plural leads to confusion with the package

docs/src/api.md
```diff
@@ -1,5 +1,167 @@
...
 
-```@autodocs
-Modules = [AbstractGPs]
-```
\ No newline at end of file
+AbstractGPs provides one main abstract type, the `AbstractGP`.
+Instances of subtypes of `AbstractGP` represent Gaussian processes -- collections of jointly-Gaussian random variables, which may be infinite-dimensional.
+AbstractGPs only become useful computationally when you specify a finite-dimensional marginal distribution to work with.
```
theogf, line 5:
```suggestion
Gaussian processes only become useful computationally when you specify a finite-dimensional marginal distribution to work with.
```
I am not sure if you meant the type here, or GPs generally

docs/src/api.md
```diff
@@ -1,5 +1,167 @@
...
+It is for this reason that we provide the `FiniteGP` type.
+Roughly speaking, this type comprises an `AbstractGP` `f` and an `AbstractVector` `x`, and represents the multivariate Gaussian distribution over `f` at `x`.
```
theogf, line 7:
```suggestion
This type comprises an `AbstractGP` : `f` and an `AbstractVector` : `x`, and represents the multivariate Gaussian distribution over `f` at every element of `x`.
```
Some kind of separation between type and instance is needed.

docs/src/api.md
```diff
@@ -1,5 +1,167 @@
...
+FiniteGP{<:MyNewGP}
+```
+We provide two ways in which to do this.
+The first is to implement methods directly on `Finite{<:MyNewGP}` -- this is detailed in the [FiniteGP APIs](@ref).
+The second is to implement some methods directly involving `MyNewGP`, and utilise default `FiniteGP` methods implemented in terms of these -- this is detailed in the [Internal AbstractGPs API](@ref).
+
+The second interface is generally easier to implement, but isn't applicable for all subtypes of `AbstractGP`.
+See [Which API should I implement?](@ref) for further discussion.
+
+Note that neither `AbstractGP` nor `FiniteGP` are exported as this package also provides user-facing concrete types which are generally what users should interact with.
```
theogf, line 22: Is this really needed?

docs/src/api.md
```diff
@@ -1,5 +1,167 @@
...
+#### Optional Methods
+Default implementations are provided for these, but you may wish to specialise for performance.
+```@docs
+mean_and_cov(::AbstractGPs.FiniteGP)
+```
+
+
+## Internal AbstractGPs API
+
+This functionality is not intended to be used directly by the users, or those building functionality on top of `AbstractGP` -- they should interact with [Primary Public API](@ref).
```
theogf, line 103: This is weird, did you mean on top of `FiniteGP` ?

willtebbutt: Indeed.

docs/src/api.md
willtebbutt: Probably not. Will remove.

docs/src/api.md
willtebbutt: Thanks :)

docs/src/api.md
willtebbutt: Sorry, could you expand on what's contradictory here?

docs/src/api.md
willtebbutt: Seems reasonable. Will ammend accordingly.

docs/src/api.md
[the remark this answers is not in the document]
willtebbutt: Hmmm yes, something like this seems to me like it would aid clarity. Where in the structure do you think that it should go?

willtebbutt:
@theogf thanks for your feedback. I think this is ready for another pass.

docs/src/api.md
```diff
@@ -1,5 +1,162 @@
...
-Modules = [AbstractGPs]
-```
\ No newline at end of file
+AbstractGPs provides one main abstract type, the `AbstractGP`.
+Instances of subtypes of `AbstractGP` represent Gaussian processes -- collections of jointly-Gaussian random variables, which may be infinite-dimensional.
+Gaussian processes only become useful computationally when you specify a finite-dimensional marginal distribution to work with.
+It is for this reason that we provide the `FiniteGP` type.
+This type comprises an `AbstractGP`, `f`, and an `AbstractVector`, `x`, and represents the multivariate Gaussian distribution over `f` at every element of `x`.
+This distribution, being finite-dimensional, is something that can be used to compute useful things.
+For example, `rand(f(x))` generates a sample from the multivariate Gaussian that is `f` at `x`, and `logpdf(f(x), y)` computes the log (marginal) probability of observing the vector `y` under this distribution.
```
theogf, line 9: The example is confusing cause we introduce `f(x)` without mentioning that this would build a FiniteGP

docs/src/api.md
```diff
@@ -1,5 +1,162 @@
...
+We provide two ways in which to do this.
+The first is to implement methods directly on `Finite{<:MyNewGP}` -- this is detailed in the [FiniteGP APIs](@ref).
+The second is to implement some methods directly involving `MyNewGP`, and utilise default `FiniteGP` methods implemented in terms of these -- this is detailed in the [Internal AbstractGPs API](@ref).
+
+The second interface is generally easier to implement, but isn't applicable for all subtypes of `AbstractGP`.
+See [Which API should I implement?](@ref) for further discussion.
+
+
+
+## Intended Audience
```
theogf, line 24: Should this go before everything else?

docs/src/api.md
theogf: Well here you say that on me "must" implement FiniteGP{<:NewGP} but this is not true since this is just one API approach (one could just implement the NewGP API

docs/src/api.md
```diff
@@ -1,5 +1,162 @@
...
+This page is intended for developers.
+If you are a user, please refer to our other examples.
+
+
+
+
+## FiniteGP APIs
+
+A `FiniteGP` represents the multivariate Gaussian induced by "indexing" into an AbstractGP `f` at an `AbstractVector` of points `x`, and adding independent zero-mean noise with covariance matrix `Σ`:
+```julia
```
theogf, line 35: Maybe define `f` first?

docs/src/api.md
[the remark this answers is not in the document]
theogf: I would put it after line 16-17 where the difference between the two API is described?

docs/src/api.md
```diff
@@ -1,5 +1,162 @@
...
+```
+
+Note that, while we _could_ provide a default implementation for `var(f, x)` as `diag(cov(f, x))`, this is generally such an inefficient fallback, that we find it preferable to error if it's not implemented than to ever hit a fallback.
+
+
+
+## Which API should I implement?
+
+To answer this question, you need to need to know whether or not the default implementations of the [FiniteGP APIs](@ref) work for your use case.
+There are a couple of reasons of which we are aware for why this might not be the case -- possibly there are others.
```
theogf, line 136: This sentence is probably not needed unless we give some concrete reasons 😊

docs/src/api.md
```diff
@@ -1,5 +1,162 @@
...
-```
\ No newline at end of file
+AbstractGPs provides one main abstract type, the `AbstractGP`.
+Instances of subtypes of `AbstractGP` represent Gaussian processes -- collections of jointly-Gaussian random variables, which may be infinite-dimensional.
+Gaussian processes only become useful computationally when you specify a finite-dimensional marginal distribution to work with.
+It is for this reason that we provide the `FiniteGP` type.
+This type comprises an `AbstractGP`, `f`, and an `AbstractVector`, `x`, and represents the multivariate Gaussian distribution over `f` at every element of `x`.
+This distribution, being finite-dimensional, is something that can be used to compute useful things.
+For example, `rand(f(x))` generates a sample from the multivariate Gaussian that is `f` at `x`, and `logpdf(f(x), y)` computes the log (marginal) probability of observing the vector `y` under this distribution.
+
```
theogf, line 10: I liked our discussion on what's an AbstractGP and what is not, concretely that an AbstractGP is an object we can use as a map. I think it would be great to have a sentence about it

theogf (commented):
OK I added/answered more comments

docs/src/api.md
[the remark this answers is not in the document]
willtebbutt: @theogf can I mark this as resolved?

docs/src/api.md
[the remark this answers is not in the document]
willtebbutt: Added. Please let me know if resolved.

docs/src/api.md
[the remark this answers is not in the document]
willtebbutt: Added. Please resolve if you're happy.

docs/src/api.md
willtebbutt: Added a clarifying sentence.

docs/src/api.md
willtebbutt: Good shout. Done.

docs/src/api.md
willtebbutt: I give the reasons below! I'll clarify that they're reasons for this.

docs/src/api.md
willtebbutt: Could you point towards a particular bit of that discussion that you found useful? I assume you're referring to #122 ?

docs/src/api.md
willtebbutt: Done.

willtebbutt:
@theogf ready for another pass.

docs/src/api.md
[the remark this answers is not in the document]
theogf: Yes sorry

docs/src/api.md
theogf: Hmm more specifically the last part of this comment : https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/issues/122#issuecomment-812063387

docs/src/api.md
theogf: I think this is the last important point which needs to be addressed.

docs/src/api.md
```diff
@@ -1,5 +1,172 @@
...
+## Introduction
+
+AbstractGPs provides one main abstract type, the `AbstractGP`.
+Instances of subtypes of `AbstractGP` represent Gaussian processes -- collections of jointly-Gaussian random variables, which may be infinite-dimensional.
+Gaussian processes only become useful computationally when you specify a finite-dimensional marginal distribution to work with.
+It is for this reason that we provide the `FiniteGP` type.
+This type comprises an `AbstractGP`, `f`, and an `AbstractVector`, `x`, and represents the multivariate Gaussian distribution over `f` at every element of `x`.
+This distribution, being finite-dimensional, is something that can be used to compute useful things.
+For example, `rand(f(x))` generates a sample from the multivariate Gaussian that is `f` at `x`, and `logpdf(f(x), y)` computes the log (marginal) probability of observing the vector `y` under this distribution.
```
theogf, line 19:
```suggestion
For example given `fx = f(x)` a `FiniteGP`, `rand(fx)` generates a sample from the multivariate Gaussian that is `f` at `x`, and `logpdf(fx, y)` computes the log (marginal) probability of observing the vector `y` under this distribution.
```
Here is in connection withe comment above

theogf (commented):
I think it looks really good now and anyway given the feedback from users we can build more upon.

The only last things which needs some concrete improvement is the part with `rand(f(x))` because it's not explained anywhere that `f(x)` produces a `FiniteGP`. In this context it looks more like an example on how to use an `AbstractGP` then the use of `FiniteGP`.

docs/src/api.md
willtebbutt: Oh, I think I might have misunderstood your comment there actually.

I'm quite sloppy with my `~` vs `=`, so possibly I've caused some confusion.

I think that an `AbstractGP` should be thought of as a distribution over functions / maps. If you could actually implement `rand(f)`, you would get a real-valued function. (Sometimes you _can_ implement `rand(f)` e.g. when the kernel is degenerate, but generally you can't).

So I think of `f(x)` as the distribution over the function at the finite collection of locations specified in `x`. Hence `rand(f(x))` gives you a vector of real numbers.

Is this in line with how you've been thinking about this?

docs/src/api.md
willtebbutt: I liked your suggestion, so I've merged it, so resolving this point.

docs/src/api.md
theogf: I mean I am in line in your understanding now.

What I think one needs to explain is why `PosteriorGP` is an `AbstractGP`. From the GP formalism it is not immediately obvious! Because the direct thought (at least mine) would be: "ah yeah the posterior! p(f|y) right? For Gaussian regression that's just a MvNormal therefore it's some kind of `FiniteGP`". So we should explain that "well this can be seen as an `AbstractGP` as well cause from this object you can also create a distribution over functions"

willtebbutt: Ahh I see. Yes, we should document this better. I'll add something about this tomorrow.

willtebbutt: @theogf I've updated the explanation to be in line with the above. Please let me know if it works for you :)

theogf (approved):

PR #131, opened 2021-04-06
Make `sampleplot` consistent with `plot`

devmotion:
This PR makes `sampleplot` consistent with `plot`. More precisely, it deprecates the current syntax `sampleplot(::FiniteGP, nsamples::Int; kwargs...)` and introduces
- `sampleplot(x::AbstractVector, gp::FiniteGP; samples=1, kwargs...)`
- `sampleplot(gp::FiniteGP; kwargs...) = sampleplot(gp.x, gp; kwargs...)`
- `sampleplot(x::AbstractVector, gp::AbstractGP; kwargs...) = sampleplot(x, gp(x, 1e-9); kwargs...)`

Here 1e-9 is used to avoid numerical issues when computing the Cholesky decomposition. The PR also fixes a weirdness/bug in the current implementation: currently we do not sample from the provided `gp::FiniteGP` but from a modified `gp.f(gp.x, 1e-9)` (due to the same numerical issues I assume). With this PR instead sampling will performed from the provided `FiniteGP`.

### DIFF
--- a/src/AbstractGPs.jl
+++ b/src/AbstractGPs.jl
@@ -28,7 +28,6 @@ export GP,
     approx_posterior,
     VFE,
     DTC,
-    sampleplot,
     update_approx_posterior,
     LatentGP,
     ColVecs,
@@ -54,4 +53,7 @@ include(joinpath("latent_gp", "latent_gp.jl"))
 
 # Plotting utilities.
 include(joinpath("util", "plotting.jl"))
+
+# Deprecations.
+include("deprecations.jl")
 end # module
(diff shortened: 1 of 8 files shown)


src/AbstractGPs.jl
```diff
@@ -28,7 +28,6 @@ export GP,
     approx_posterior,
     VFE,
     DTC,
-    sampleplot,
```
devmotion, line 31: This is not needed, `@userplot` exports both `sampleplot` and `sampleplot!` automatically.

willtebbutt (approved):
I'm happy with this.

PR #133, opened 2021-04-08
Build stable docs

devmotion:
Currently new tags do not trigger the Docs action and therefore the stable docs are not updated.


willtebbutt (approved):
I'm definitely not qualified to review this, but I'm going to assume that this will fix it 😂

devmotion:
It's the same configuration as in KernelFunctions, and there the stable docs are built correctly :slightly_smiling_face:

PR #62, opened 2020-09-25
Soss test

cscherrer:
This PR adds a test for Soss. 

This doesn't work yet, because Turing and Soss have some name conflicts (at least `@model`). What do you think @theogf , I guess the tests for these systems could be in different modules? Or maybe just `import` instead of `using` and qualify all the names?

PR #135, opened 2021-04-12
Soss test (update of #62)

devmotion:
I just updated https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/pull/62 some days ago. However, unfortunately, sampling with Soss fails due to errors such as
```julia
  Test threw exception
  Expression: length(Soss.sample(DynamicHMCChain, m | (y = y,), 5)) == 5
  UndefVarError: testvalue not defined
  Stacktrace:
    [1] getproperty
      @ ./Base.jl:26 [inlined]
    [2] macro expansion
      @ ~/.julia/packages/GeneralizedGenerated/PV9u7/src/closure_conv.jl:132 [inlined]
    [3] _xform(M::Type{GeneralizedGenerated.NGG.TypeLevel{Module, "Buf{17}()"}}, _m::Soss.Model{NamedTule{(:X,), T} where T<:Tuple, GeneralizedGenerated.NGG.TypeLevel{Expr, "Buf{23}()"}, GeneralizedGenerate.NGG.TypeLevel{Module, "Buf{17}()"}}, _args::NamedTuple{(:X,), Tuple{Vector{Vector{Float64}}}}, _data::amedTuple{(:y,), Tuple{Vector{Int64}}})
      @ Soss ~/.julia/packages/GeneralizedGenerated/PV9u7/src/closure_conv.jl:132
    [4] xform(m::Soss.ConditionalModel{NamedTuple{(:X,), T} where T<:Tuple, GeneralizedGenerated.NGG.TyeLevel{Expr, "Buf{23}()"}, GeneralizedGenerated.NGG.TypeLevel{Module, "Buf{17}()"}, NamedTuple{(:X,), Tple{Vector{Vector{Float64}}}}, NamedTuple{(:y,), Tuple{Vector{Int64}}}})
      @ Soss ~/.julia/packages/Soss/PRbUF/src/primitives/xform.jl:23
    [5] sample(rng::Random._GLOBAL_RNG, ::Type{DynamicHMCChain}, m::Soss.ConditionalModel{NamedTuple{(:,), T} where T<:Tuple, GeneralizedGenerated.NGG.TypeLevel{Expr, "Buf{23}()"}, GeneralizedGenerated.NGG.ypeLevel{Module, "Buf{17}()"}, NamedTuple{(:X,), Tuple{Vector{Vector{Float64}}}}, NamedTuple{(:y,), Tupe{Vector{Int64}}}}, nsamples::Int64, nchains::Int64)
      @ Soss ~/.julia/packages/Soss/PRbUF/src/samplechains/dynamichmc.jl:15
    [6] sample (repeats 2 times)
      @ ~/.julia/packages/Soss/PRbUF/src/samplechains/dynamichmc.jl:29 [inlined]
    [7] macro expansion
      @ ~/.julia/dev/AbstractGPs/test/compat/soss.jl:49 [inlined]
    [8] macro expansion
      @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/Test/src/Test.jl:1151 [inined]
    [9] macro expansion
      @ ~/.julia/dev/AbstractGPs/test/compat/soss.jl:31 [inlined]
   [10] macro expansion
      @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/Test/src/Test.jl:1151 [inined]
   [11] top-level scope
      @ ~/.julia/dev/AbstractGPs/test/compat/soss.jl:2
```

Since I am new to Soss I am not completely sure but to me it seems this could be a bug in Soss? Or do I use it incorrectly @cscherrer?


test/runtests.jl
```diff
@@ -23,17 +23,19 @@ using AbstractGPs:
 
 using Documenter
 using ChainRulesCore
-using Distributions: MvNormal, PDMat
+using Distributions: MvNormal, PDMat, Poisson, LogNormal, sample
```
devmotion, line 26:
```suggestion
using Distributions: MvNormal, PDMat, Poisson, LogNormal
```

cscherrer:
Thanks @devmotion and @willtebbutt ! I'm sure there's just a missing method, and I'd love to have some more tests anyway. I'll have a look

devmotion:
An additional problem is that apparently SampleChainsDynamicHMC requires Julia 1.5.

cscherrer:
**testvalue**

This is defined in MeasureTheory. The semantics are to return a point in the support, like `rand` without the randomness. I use this in the process of determining the transform for the model. `rand` adds a little overhead, and requires a `rand` method that some measures don't have. Does Distributions have something like this? If not I could define
```julia
MeasureTheory.testvalue(d::Distributions.Distribution) = rand(d)
```

**Julia 1.5**
I'm not sure I need this in SampleChains etc, mostly I figured there's not much reason to not be using at least 1.5. I do require it in MeasureTheory for the `(;x)` trick it adds.

cscherrer:
I didn't realize I'm reexporting MeasureTheory from Soss, I'll need to change that to make Distributions usable. This is a breaking change, so we'll be at v0.18

devmotion:
> I'm not sure I need this in SampleChains etc, mostly I figured there's not much reason to not be using at least 1.5.

This is quite unfortunate since there is no reason for us to drop support for Julia 1.3 and 1.4 at this point. Probably then we have to separate Soss and its dependencies into a separate environment if we want to test it, and make the tests optional.

cscherrer:
I'll try things out with 1.3, maybe it works or could be made to work without much trouble

test/compat/soss.jl
```diff
@@ -0,0 +1,51 @@
...
+            kernel = α * transform(SqExponentialKernel(), 1 / ρ)
+            f = GP(kernel)
+
+            # Sampling Distribution.
+            y ~ f(X, σ²)
+        end
+
+        # Test for matrices
+        m = gp_regression(; X=RowVecs(X))
+        @test length(Soss.sample(DynamicHMCChain, (m | (y=y,)), 5)) == 5
```
cscherrer, line 24:
```suggestion
        @test length(Soss.sample(DynamicHMCChain, (m | (y=y,)), 5, 1)) == 5
```

cscherrer:
Getting closer! I'm now at
```julia
julia> include("test/compat/soss.jl")
latent GP regression: Error During Test at /home/chad/git/AbstractGPs.jl/test/compat/soss.jl:45
  Test threw exception
  Expression: length(Soss.sample(DynamicHMCChain, m | (y = y,), 5)) == 5
  Not implemented:
  xform(FiniteGP{GP{ZeroMean{Float64}, Matern32Kernel}, RowVecs{Float64, Matrix{Float64}, SubArray{Float64, 1, Matrix{Float64}, Tuple{Int64, Base.Slice{Base.OneTo{Int64}}}, true}}, Diagonal{Float64, Fill{Float64, 1, Tuple{Base.OneTo{Int64}}}}}(
  f: GP{ZeroMean{Float64}, Matern32Kernel}(ZeroMean{Float64}(), Matern 3/2 Kernel)
  x: SubArray{Float64, 1, Matrix{Float64}, Tuple{Int64, Base.Slice{Base.OneTo{Int64}}}, true}[[1.284670749722735], [-0.48905837672741387], [1.234744953422099]]
  Σy: [1.0e-18 0.0 0.0; 0.0 1.0e-18 0.0; 0.0 0.0 1.0e-18]
  )
  )
```

This means Soss needs to know how to transform a Gaussian process to R^n, so we'll need to think about the right way to set this up. The obvious approach is just to make AbstractGPs a dependency of Soss (probably with `@require`, but I'd rather find a simpler way if there is one. I'm mostly using TransformVariables for this.

devmotion:
> This means Soss needs to know how to transform a Gaussian process to R^n, so we'll need to think about the right way to set this up. The obvious approach is just to make AbstractGPs a dependency of Soss (probably with @require, but I'd rather find a simpler way if there is one. I'm mostly using TransformVariables for this.

IMO it should just work since `FiniteGP` is a subtype of `Distributions.AbstractMvNormal`. There is no need to depend on AbstractGPs.

devmotion:
Regarding your previous question:

> Does Distributions have something like this?

No, there's nothing like this in Distributions (at least I am not aware of anything). Probably the safest option is to sample from the distribution.

willtebbutt (approved):
This is great. I just have one question that I don't want to prevent you from merging.

Regarding the `GROUP` environment variable, do we ever set it to anything other than `All` or `AbsrtactGPs`? I'm just wondering whether there are actually situation in which we just want to run the compat tests?

.github/workflows/ci.yml
```diff
@@ -40,6 +40,9 @@ jobs:
             ${{ runner.os }}-
       - uses: julia-actions/julia-buildpkg@v1
       - uses: julia-actions/julia-runtest@v1
+        env:
+          # run compat tests with Turing and Soss only with stable Julia version
+          GROUP: ${{ matrix.version == '1' && 'All' || 'AbstractGPs' }}
```
theogf, line 45: Can we have parenthesis to avoid ambiguities here?

devmotion:
No, probably the "Compat" alternative could/should be removed.

Some background info:
Initially, I had only added a check of the Julia version (>= v"1.5") to avoid test failures on older versions that are not supported by Soss anymore while still being able to run our standard tests on Julia 1.3. However, I noticed that one also has to upper bound the Julia version - Turing depends on Libtask_jll which has to be built against libjulia which does not exist in Yggdrasil for unstable Julia versions. Due to this issue tests will fail on nightly (it works currently in the regular test setup since Pkg pulls in an old Turing version that did not depend on Libtask_jll, in the separate project environment it is resolved differently). To fix these issues one could check for Julia 1.6 explicitly (== v"1.6") but then one would have to update the check manually for new stable versions (otherwise these tests are just skipped). I introduced the environment variable to ensure that Soss and Turing are always tested with the stable Julia version (the version check is included to prevent anyone, also offline, from trying to run the tests with older Julia versions).

devmotion:
It also seems that sometimes Soss/DynamicHMC fails due to numerical errors. I assume that the variance can become too small, either in the warmup phase or during sampling. I'll check if it is sufficient to add some small positive offset (similar to what we do when plotting samples).

.github/workflows/ci.yml
devmotion: I added parentheses :+1:

PR #139, opened 2021-04-15
Remove `struct`s in the example

devmotion:
This PR addresses https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/pull/106#discussion_r577900904 and replaces the callable structs with regular functions. I still think we should avoid closures with global variables in our example since in practice it often leads to terrible performance if the global variables are not declared as constant. The implementation could be simplified a bit more by making the variables constant, but I think it's annoying to annotate all these variables and I assume it's easy to miss their importance, in particular if you are new to Julia.

### DIFF
--- a/examples/regression_1d.jl
+++ b/examples/regression_1d.jl
@@ -106,7 +106,8 @@ plot!(0:0.001:1, p_fx; label=false)
 
 function gp_loglikelihood(x, y)
     function loglikelihood(params)
-        kernel = softplus(params[1]) * (Matern52Kernel() ∘ ScaleTransform(softplus(params[2])))
+        kernel =
+            softplus(params[1]) * (Matern52Kernel() ∘ ScaleTransform(softplus(params[2])))
         f = GP(kernel)
         fx = f(x, 0.1)
         return logpdf(fx, y)
@@ -235,7 +236,9 @@ using LogDensityProblems
 # We have to implement the LogDensityProblems interface for `loglik_train`.
 
 ## Log joint density
-LogDensityProblems.logdensity(ℓ::typeof(loglik_train), params) = ℓ(params) + logprior(params)
+function LogDensityProblems.logdensity(ℓ::typeof(loglik_train), params)
+    return ℓ(params) + logprior(params)
+end
 
 ## The parameter space is two-dimensional
 LogDensityProblems.dimension(::typeof(loglik_train)) = 2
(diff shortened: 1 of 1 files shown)


devmotion:
@4aHxKzD Do you think the version in this PR without callable structs (https://juliagaussianprocesses.github.io/AbstractGPs.jl/previews/PR139/examples/regression_1d/) is easier to understand and addresses your concerns?

4aHxKzD:
> @4aHxKzD Do you think the version in this PR without callable structs (https://juliagaussianprocesses.github.io/AbstractGPs.jl/previews/PR139/examples/regression_1d/) is easier to understand and addresses your concerns?

I would say yes, but it is possible that I have some bias since at the time I had already encountered the concept of functions that return another function, but not the concept of callable structs. Today both make sense too me.

Maybe this way it looks more like code that is not as much idiomatic Julia, so a Python user that tries to read this example would have an easier time compared to the previous version.

examples/regression_1d.jl
```diff
@@ -241,17 +233,19 @@ plt
 using DynamicHMC
 using LogDensityProblems
```
theogf, line 234: I know it was there already but do we really need to complicate things by using an extra dependency?
I agree it's probably better but it makes it also harder to understand.

devmotion: Do you suggest removing the DynamicHMC example? If we were to remove one approach, I'd remove the AdvancedHMC example - IMO it is less clear and difficult/impossible to understand without knowing the details of AdvancedHMC. In any case, I think this should go into a separate PR :slightly_smiling_face:

theogf: Ah sorry I mean the `LogDensityProblems` package.

devmotion: It is required to define the necessary overloads for DynamicHMC: https://tamaspapp.eu/DynamicHMC.jl/dev/worked_example/

devmotion: And this functionality is not (re)exported by DynamicHMC.

theogf: Ah sorry I did not know it was needed for `DynamicHMC`. Then maybe this should be left as a quick comment?

theogf:
```suggestion
using LogDensityProblems # Interface to work with DynamicHMC
```

devmotion: Below this line it says: "We have to implement the LogDensityProblems interface for loglik_train." Is that sufficient?

theogf: Not really since it does not explain at all why we need to implement the `LogDensityProblems` interface in the first place!

devmotion: I moved the comment and made it more explicit. Is it better now?

devmotion:
@theogf Can you approve if you are satisfied with the PR?

theogf (approved):

PR #141, opened 2021-04-15
Simplify `map` in example

devmotion:
This PR contains a tiny change: it replaces `[map(f, x) for x in xs]` with `map(f, xs)` in the example.

### DIFF
--- a/examples/regression_1d.jl
+++ b/examples/regression_1d.jl
@@ -179,7 +179,7 @@ samples, _ = sample(
 # We transform the samples back to the constrained space and compute the mean of both
 # parameters:
 
-samples_constrained = [map(softplus, p) for p in samples]
+samples_constrained = map(softplus, samples)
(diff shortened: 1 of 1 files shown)


theogf (approved):

willtebbutt (approved):

devmotion:
Haha I didn't realize that of course it is implemented in such a way since `x` is a vector. With Julia 1.6, one could write `map(.f, xs)` instead but I think probably in the example one should not use syntax that is only available on Julia 1.6.

theogf:
Can't you do `map(x -> f.(x), xs)` then?

devmotion:
Yes, or `map(x -> map(f, x), xs)` or `map(Base.Fix1(map, f), xs)`. But it's not as simple as I thought it could be and only slightly simpler than the current implementation at most :disappointed:

PR #149, opened 2021-05-03
Bump Some Versions

willtebbutt:


devmotion:
Seems to be a duplicate of https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/pull/145 :slightly_smiling_face: The main difference is that in the other PR docs/Manifest.toml is updated as well.

willtebbutt:
Right you are!

Issue #161, opened 2021-06-03
Updating Dependencies


willtebbutt:
We're currently going through a ChainRulesCore breaking change. AbstractGPs has to wait for both Turing.jl and Soss.jl to update as well before it's possible for us to update -- they're fantastic packages of course, but they're dependency-heavy so tend to take a little while to update everything.

I definitely want to ensure that we're testing against them, but I wonder whether it would make sense to have (at least) two separate CI runs -- one for the core functionality of AbstractGPs, and another to run Turing and Soss (or possibly separate runners for Turing and Soss?).

This would mean that we'd be able to plough ahead with upgrades for AbstractGPs provided that our "core" tests pass, but we would be kept apprised of whether or not AbstractGPs interacts properly with Soss and Turing.

Thoughts?

devmotion:
I think this would make sense, we already separated them into a separate environment since Soss is only compatible with Julia >= 1.5.

devmotion:
So it would be a trivial change, we basically just have introduce a new "GROUP" value (maybe "PPL"?) and add an action that runs it.

PR #162, opened 2021-06-03
Reorganize tests and update CI

devmotion:
This PR separates the tests of Turing + Soss a bit more and fixes https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/issues/161. I moved them to a separate action such that CI tests are still shown as passing, also in the README, if the Turing + Soss tests fail :slightly_smiling_face: 

I also thought it's sufficient to run the PPL tests with the stable Julia version on Linux.


willtebbutt (approved):
This is great. Please feel free to merge when CI passes.

devmotion:
BTW tests with Julia nightlies are broken since `isgenerated` was renamed to `hasgenerator` on the master branch but IRTools uses `isgenerated`. It seems it would be fixed by https://github.com/FluxML/IRTools.jl/pull/86 (or by reverting the change in base or at least re-adding `isgenerated`, there's some discussion in the original PR it seems).

PR #167, opened 2021-06-04
Update Turing and Soss dependencies

devmotion:


willtebbutt (approved):
Looks like Distibutions@0.25 still isn't being used, but Turing, Soss, and SampleChainsDynamicsHMC have been updated, so I'm going to approve.

theogf:
Since we just want to maintain compatibility with Turing and Soss do we really need to check `nightly` ?

devmotion:
Only the standard CI uses nightly, Turing and Soss are only tested with the stable Julia version and only on linux.

devmotion:
> Looks like Distibutions@0.25 still isn't being used, but Turing, Soss, and SampleChainsDynamicsHMC have been updated, so I'm going to approve.

Yes, I noticed this. I thought also we should drop old versions (Turing < 0.16 is not compatible with ChainRulesCore 0.10), so maybe let's stick with Distributions 0.24 for now and use the CompatHelper PR as a reminder that it should be updated eventually?

theogf:
> Only the standard CI uses nightly, Turing and Soss are only tested with the stable Julia version and only on linux.

Oh right I read the CI wrong! :sweat_smile:

devmotion:
Good to go?

theogf (approved):

Issue #173, opened 2021-06-15
plot() recipe returns unexpected prediction intervals


azev77:
```julia
# Load dataset.
using HTTP, CSV, DataFrames, Dates
link = "https://bit.ly/3vtTaW5";
r = HTTP.get(link);
df = CSV.read(r.body, DataFrame; missingstring="NA");

using AbstractGPs, Plots
X = 0:1.0:(length(df.date)-1) 
Y = df.value;
XH = 0:1.0:(length(df.date)-1 +5) 

f = GP(Matern32Kernel()) # Define GP prior w Matern32 kernel
fx = f(X, 0.001)         # Finite projection at the inputs `X` 
logpdf(fx, Y)            # -512 # Data's log-likelihood w.r.t prior GP `f`. 
p_fx = posterior(fx, Y)  # Exact posterior given `Y`.
logpdf(p_fx(X), Y)       # 184 # Data's LL w.r.t prior posterior GP `p_fx`.

px = XH[end-4:end]
pm = mean.(marginals(p_fx(XH))[end-4:end])
ps = std.(marginals(p_fx(XH))[end-4:end])

# Plot posterior.
scatter(X, Y; label="Data")
plot!(XH, f(XH, 0.001); label="Prior")
plot!(p_fx(XH), label="Posterior" )
scatter!(px, pm, lab = "oos prediction μ")
scatter!(px, pm - 1.0*ps, lab="oos prediction μ - σ") 
scatter!(px, pm + 1.0*ps, lab="oos prediction μ + σ") 
```
![image](https://user-images.githubusercontent.com/7883904/122110865-0a60ee80-cded-11eb-9f93-9f2862422eb5.png)
It looks like the plot() recipe returns `μ ± 1.0σ` instead of (what I'd expect) `μ ± 1.96σ`.
Is this intended? Can this be customized?

azev77:
Looks like it might be set to 1.0 here:
https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/blob/0e223c8cd57d625c9476f9e4365beaf80b7fad14/src/util/plotting.jl#L6

Not sure & idk how to modify it...

theogf:
The 1.96sigma  corresponds to the 95%quantile but there is no particular reason to prefer it to just 1sigma or 2 sigma.
What would you expect 1.96 in particular?

azev77:
>  What would you expect 1.96 in particular?
Not sure what you mean. 

[...]

PR #174, opened 2021-06-16
Add documentation about the plotting

theogf:
Adresses #173 for the lack of documentation on plotting

### DIFF
--- a/docs/src/concrete_features.md
+++ b/docs/src/concrete_features.md
@@ -103,5 +103,11 @@ You can directly plot your GP prediction via [Plots.jl](https://github.com/Julia
 We provide two functions `plot` and `sampleplot` taking as arguments `X, AbstractGP` or `FiniteGP`
 For example uses see the [One-dimensional regression](@ref) tutorial
 
+```@docs
+plot(::FiniteGP)
+sampleplot
+```
+
+
 ##### Makie.jl
 For using `Makie.jl` you can use the additional package [AbstractGPsMakie.jl](https://github.com/JuliaGaussianProcesses/AbstractGPsMakie.jl).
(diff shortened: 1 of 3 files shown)


devmotion:
Could we maybe include the existing docstring of `sampleplot` (instead)? And add a docstring for `plot`?

theogf:
Ok now I see it's quite overlapping with https://juliagaussianprocesses.github.io/AbstractGPs.jl/previews/PR174/examples/regression_1d/

How about I just put the docstrings in `Concrete Features` and that in the `Regression example` I add some of the kwargs (`ribbon_scale` and `samples`)?

theogf:
@devmotion Do you know where `x` is  sorted before being used? I cannot find where it happens. It's nice but I am annoyed to not know where it's from

devmotion:
AFAIK for some plot types Plots sorts stuff internally before plotting it.

azev77:
Guys, the autoplot() function in forecast.R as a default plots 80% & 95% prediction intervals

https://pkg.robjhyndman.com/forecast/reference/plot.forecast.html

devmotion:
That seems quite arbitrary :shrug: IMO this is more a documentation issue (which this PR addresses) than a problem with the default values.

azev77:
But how is the current default choice of `μ ± 1.0σ` any less arbitrary than the more commonly used `μ ± 1.96σ`?

willtebbutt:
I don't think it's more arbitrary either way. I personally have no strong feelings about it, but if we were change the defaults, we'd have to do it in the next breaking release. I can see the appeal of plotting 95% though 🤷

theogf:
@devmotion I feel a bit dumb but I have no idea how to include the `plot` docstring...

PR #176, opened 2021-06-22
Support and test in-place sampling

devmotion:
Currently something like
```julia
julia> gp = GP(SqExponentialKernel());

julia> x = rand(11);

julia> rand!(f(x), similar(y))
```
throws an error since `Distributions._rand!` is not defined for `FiniteGP`s (this is the method that Distributions falls back to when one calls `rand!` for multivariate distributions and that one is supposed to define for new multivariate distributions: https://juliastats.org/Distributions.jl/latest/extends/#Multivariate-Sampler).

This PR adds a definition of `Distributions._rand!` for `FiniteGP`s. Distributions also falls back to it when calling `rand` but I think often this is a bad idea (one has to use type heuristics that fail sometimes - e.g. tests fail if one removes the custom definitions in AbstractGPs) and so I did not remove or modify the existing definitions of `rand`. It might still be useful to define `Distributions._rand!` instead of `Random.rand!` since then e.g. Distributions already checks if the size of the input is correct and one gets support for arrays of vectors for free (see https://github.com/JuliaStats/Distributions.jl/blob/a0cb0969d755872586ec79e985503f03beae9d08/src/multivariates.jl#L47-L66).

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -64,7 +64,7 @@ If you are building something on top of AbstractGPs, try to implement it in term
 
 ```@docs
 rand
-rand!
+Random.rand!
 marginals
 logpdf(::AbstractGPs.FiniteGP, ::AbstractVector{<:Real})
 posterior(::AbstractGPs.FiniteGP, ::AbstractVector{<:Real})
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -64,15 +64,16 @@ If you are building something on top of AbstractGPs, try to implement it in term
 
 ```@docs
 rand
+rand!
 marginals
 logpdf(::AbstractGPs.FiniteGP, ::AbstractVector{<:Real})
 posterior(::AbstractGPs.FiniteGP, ::AbstractVector{<:Real})
 mean(::AbstractGPs.FiniteGP)
 var(::AbstractGPs.FiniteGP)
-
 ```
 
 #### Optional methods
+
 Default implementations are provided for these, but you may wish to specialise for performance.
 ```@docs
 mean_and_var(::AbstractGPs.FiniteGP)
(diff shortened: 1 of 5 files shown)


docs/src/api.md
```diff
@@ -64,15 +64,16 @@ If you are building something on top of AbstractGPs, try to implement it in term
 
 ```@docs
 rand
+rand!
```
theogf, line 67: Should we actually replace this by `Distributions._rand!` instead?

devmotion: I guess this would be reasonable if we want to follow the conventions in Distributions. I am just not sure if we want to do this, it feels weird if an internal method such as `_rand!` is part of the API (of course, this is a problem of Distributions but since we already define `rand` separately maybe we want to not follow it here either).

theogf: >  if we want to follow the conventions in Distributions

That seems like something we want to do no? Maybe we can add a reference to their API docs for better understanding?

devmotion: But if we really want to follow their conventions we would/should not have to implement `rand`. However, I really think we should not do this since it is quite likely to lead to incorrect types (I checked and tests fail since the output type is incorrect for `Float32` - of course, one could work around this by defining `eltype(::FiniteGP)` but I think the easier solution is to only perform in-place sampling if requested :shrug:). So we do not follow the abstractions in Distributions even if we tell users to implement `Distributions._rand!`.

theogf (approved):

docs/src/api.md
willtebbutt: The other reason to like having an explicit implementation for `rand` is AD, since our AD tools really don't like mutation.

willtebbutt (approved):
I'm also happy with this.

PR #185, opened 2021-07-19
move to KernelFunctions style literate.jl

st--:
Copying over the new notebooks setup from KernelFunctions.jl with independent environments for each example.


docs/Project.toml
```diff
@@ -1,28 +1,8 @@
...
-EllipticalSliceSampling = "cad2338a-1db2-11e9-3401-43bc07c9ede2"
-ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
-Literate = "98b081ad-f1c9-55d3-8b20-4c87d4299306"
-LogDensityProblems = "6fdf6af0-433a-55f7-b3ed-c6c6e0b8df7c"
-Optim = "429524aa-4258-5aef-a3af-852621145aeb"
-Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
-StatsFuns = "4c63d2b9-4356-54db-8cca-17b64c39e42c"
 
 [compat]
```
st--, line 6: NB- KernelFunctions.jl includes a line `julia = "1.3"`, do we want to add a line like that here, or remove it there, and should it be 1.3 or just `julia = "1"`, or `julia = "1.6"`?

docs/literate.jl
```diff
@@ -0,0 +1,53 @@
...
+#
+#md # [![](https://img.shields.io/badge/show-nbviewer-579ACA.svg)](@__NBVIEWER_ROOT_URL__/examples/@__NAME__.ipynb)
+#md #
+# *You are seeing the
+#md # HTML output generated by [Documenter.jl](https://github.com/JuliaDocs/Documenter.jl) and
+#nb # notebook output generated by
+# [Literate.jl](https://github.com/fredrikekre/Literate.jl) from the
+# [Julia source file](@__REPO_ROOT_URL__/examples/@__NAME__/script.jl).
+#md # The corresponding notebook can be viewed in [nbviewer](@__NBVIEWER_ROOT_URL__/examples/@__NAME__.ipynb).*
+#nb # The rendered HTML can be viewed [in the docs](https://juliagaussianprocesses.github.io/AbstractGPs.jl/dev/examples/@__NAME__/).*
```
st--, line 34: The one line changed from KernelFunctions.jl

docs/make.jl
```diff
@@ -1,51 +1,81 @@
-using Documenter
+### Process examples
+# Always rerun examples
+const EXAMPLES_OUT = joinpath(@__DIR__, "src", "examples")
+ispath(EXAMPLES_OUT) && rm(EXAMPLES_OUT; recursive=true)
+mkpath(EXAMPLES_OUT)
 
-# Print `@debug` statements (https://github.com/JuliaDocs/Documenter.jl/issues/955)
-if haskey(ENV, "GITHUB_ACTIONS")
-    ENV["JULIA_DEBUG"] = "Documenter"
```
st--, line 5: Now handled by .github/workflows/docs.yml

docs/make.jl
```diff
@@ -1,51 +1,81 @@
...
+    end;  # we have to load all packages used (implicitly) within jldoctest blocks in the API docstrings
+    recursive=true,
 )
 
 makedocs(;
-    modules=[AbstractGPs],
-    format=Documenter.HTML(),
-    repo="https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/blob/{commit}{path}#L{line}",
     sitename="AbstractGPs.jl",
+    repo="https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/blob/{commit}{path}#L{line}",
```
st--, line 60: Do we want to add a `repo = "..."` to KernelFunctions.jl's docs/make.jl?

docs/make.jl
```diff
@@ -1,51 +1,81 @@
...
+            map(filter!(filename -> endswith(filename, ".md"), readdir(EXAMPLES_OUT))) do x
+                return joinpath("examples", x)
+            end,
+    ],
+    strict=true,
+    checkdocs=:exports,
+    doctestfilters=[
+        r"{([a-zA-Z0-9]+,\s?)+[a-zA-Z0-9]+}",
+        r"(Array{[a-zA-Z0-9]+,\s?1}|Vector{[a-zA-Z0-9]+})",
+        r"(Array{[a-zA-Z0-9]+,\s?2}|Matrix{[a-zA-Z0-9]+})",
```
st--, line 77: Let's see if this works

docs/make.jl
willtebbutt: I'm not sure what is considered best-practice here unfortunately. Is this something you've seen done elsewhere?

Interestingly, we don't appear to provide the `repo` kwarg in KernelFunctions.jl -- according to the docstring for `makedocs`, since we're using github it should be able to figure this out automatically, so perhaps we can just remove this kwarg?

examples/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+Every example uses a separate project environment. Therefore you should also create a new
+project environment in the directory of the example that contains all packages required by your script.
+Note that the dependencies of your example *must* include the `Literate` package.
+
+From a Julia REPL started in your example script's directory, you can run
+```julia
+julia> ] activate .
+julia> ] add Literate
+julia> ] add AbstractGPs
```
willtebbutt, line 47: Do we want to do this, or to say `dev ../..` to ensure that they're getting the branch of AbstractGPs on which they're working? Possibly this is probably desirable for testing purposes, as it means that we can tie the examples in with our CI to ensure that they don't go out of date (I've done this in TemporalGPs).

willtebbutt: I've just had a look at what we do in KernelFunctions, and this doesn't appear to be what we currently do there. I wonder whether we should adopt this? I've opened https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/339 to discuss.

In any case, I'm happy for this to go in without changing this, so please disregard this comment if you don't want to address it in this PR.

willtebbutt (commented):
This is fantastic -- thanks for sorting it out.

The Manifest for the example needs updating -- presently it assumes that AbstractGPs is located at `..` -- the fix is just to `dev ../..` to update the Manifest.

Additionally, the optimisation example at the end of the script seems to be a bit unstable -- I was getting cholesky errors when I ran it locally. Would you mind adding some jitter to the pseudo-points while we're here? Specifically, making line 395 something like
```julia
        return -elbo(fx, y, f(logistic.(params[3:end]), 1e-6))
```
fixes the problem for me.

docs/make.jl
st--: so in this file this isn't actually a change (it just moved down a line), I'm just asking cos KernelFunctions.jl does not provide the repo kwarg and so there's an inconsistency between the two setups - I think we should resolve that, but have no idea/opinions on which direction:)

st--:
```suggestion
```
then ?

examples/README.md
st--: Oh yeah, it was broken but that was because it had been `dev ..` - happy to change it to `dev ../..` instead of pinning (probably what we want!)

docs/make.jl
willtebbutt: Yeah -- let's do that.

examples/README.md
willtebbutt: Great -- turns out we're just a bit inconsistent in KernelFunctions anyway, so I'll make a PR to make everything `dev ../..`

docs/Project.toml
theogf: It should be 1.3 for the main project.toml. It means we do not support lower versions of Julia.
For the docs we could add it, but it's not particularly necessary.

willtebbutt: I would personally prefer to not include a julia bound, and let the examples inherit it from whatever the bound the bound is on the current version of AbstractGPs.

examples/README.md
willtebbutt: > Possibly this is probably desirable for testing purposes, as it means that we can tie the examples in with our CI to ensure that they don't go out of date (I've done this in TemporalGPs).

I wasn't thinking properly when I wrote this -- this PR of course does this when the docs are built.

examples/README.md
```diff
@@ -0,0 +1,52 @@
+# Example notebooks
+
+The examples in this directory are stored in [Literate.jl](https://github.com/fredrikekre/Literate.jl) format.
+
+To run them locally, navigate to the directory with the example that you want to run and
```
willtebbutt, line 5: I wonder whether we should just put something like
```julia
using Pkg
Pkg.activate(@__DIR__)
Pkg.resolve()
Pkg.instantiate()
```
at the top of each `script.jl` and tell users just to run the script from wherever? This would both simplify the instructions, and ensure that the manifest has the correct version of AbstractGPs in it (I'm not sure whether it actually matters though)

examples/README.md
st--: changed to `dev ../..`

examples/README.md
st--: Running the `resolve()` means the Manifest.toml would change on the fly so we should then take that out from the repo. That would mean *not* tying any specific versions, right? Not sure if that's what we want.. let's leave it for now?

willtebbutt: Fair point -- I agree we should leave it for now.

willtebbutt (approved):
I'm happy with this now. With our current settings I don't believe you'll be able to merge until we resolve the broken AD stuff though, which is annoying.

edit: I was getting confused and thought this was in KernelFunctions. Ignore my AD-related comments.

Issue #193, opened 2021-07-30
Approximate Inference


willtebbutt:
Currently, we have
```julia
approx_posterior(approximation, fx, y, u)
```
As pointed out by @st-- and @rossviljoen in [SparseGPs.jl](https://github.com/rossviljoen/SparseGPs.jl/issues/13), we should consider reducing this to a 3-arg function in which `approximation` contains `u`, since `u` is really a component of the approximation. This kind of thing would generalise more elegantly to what @rossviljoen is doing in `SparseGPs.jl`, as it will be natural in that case to put the variational parameters associated with `q(u)` inside `approximation` as well. More generally, there are approximate inference algorithms which don't involve pseudo-points, and it would be nice to generalise to them also.

So the new `approx_posterior` function for the saturated `VFE` approximation would be something like
```julia
approx_posterior(VFE(u), fx, y)
```
and for the unsaturated would be something like
```julia
approx_posterior(VFE(u, qu), fx, y)
```
(or something a bit like that).

willtebbutt:
I wonder whether we actually just want to keep the same function names and always pass an `approximation` argument?

For example,
```julia
posterior(approximation, fx, y)
logpdf(approximation, fx, y)
rand(rng, approximation, fx)
```
etc?

st--:
@rossviljoen @willtebbutt is this sufficiently resolved by #194 ?

st--:
Reviving this discussion, another question is where should the computation/optimisation go? E.g. for SparseVariationalApproximation we need to optimise to find the optimal q(u), and currently this is left as an exercise for the user, and `posterior(sva, lfx, y)` is then super fast (and ignores the last two arguments entirely). In contrast, `posterior(LaplaceApproximation(), lfx, y)` actually computes the mode of the posterior. From a user perspective, it'd be nice if it was consistent (and easy...).

willtebbutt:
Yeah -- I at least think we ought to be able to do this in the deterministic-objective case (no minibatching, quadrature / exact reconstruction-term computation) at the very least. I've done something like this in `ConjugateComputationVI.jl` (which I should move over to here and align with what you've done with the Laplace approximation).

willtebbutt:
I believe that this is now stale, so am closing.

PR #194, opened 2021-08-04
Refactor Approximate Inference

rossviljoen:
Implements the idea in #193 (see also https://github.com/rossviljoen/SparseGPs.jl/issues/13)

Essentially, this replaces the current method for constructing approximate posteriors:
```approx_posterior(VFE(), fx, y, fz)```
with
```posterior(VFE(z), fx, y)```

(n.b. dropping the `approx_` prefix - should we keep it?)

The reasoning is that using this style of approximation generalises more easily to cases where you need more complex approximations such as in [SparseGPs](https://github.com/rossviljoen/SparseGPs.jl) - e.g. using a variational distribution `q`: ```posterior(SVGP(z, q), fx, y)```. 

Furthermore, it only requires passing the inducing inputs `z` instead of a full `fz isa FiniteGP` which enforces the fact that `fz` and `fx` are constructed from the same underlying process `fx.f`. 

It also moves the elbo and dtc from `finite_gp.jl` to `approx_posterior_gp.jl` and redefines them from:

```elbo(fx, y, fz)```
to
```elbo(VFE(z), fx, y)```

---

As @willtebbutt mentioned in #193 - it would be useful to define other functions like `logpdf` and `rand` on the approximation:

```
posterior(approximation, fx, y)
logpdf(approximation, fx, y)
rand(rng, approximation, fx)
```

### CODE REVIEW

File: src/posterior_gp/approx_posterior_gp.jl [hunk 1]
```diff
@@ -1,4 +1,25 @@
...
+struct VFE{Tz<:AbstractVector,Tj<:Real}
+    z::Tz
+    jitter::Tj
+end
+
+const default_jitter = default_σ²
+
+function VFE(z::AbstractVector, jitter=default_jitter)
+    return VFE(z, jitter)
+end
```
willtebbutt, line 21:
```suggestion
VFE(z::AbstractVector; jitter=default_jitter) = VFE(z, jitter)
```

src/posterior_gp/approx_posterior_gp.jl
```diff
@@ -1,4 +1,25 @@
-struct VFE end
+"""
+    VFE{Tz<:AbstractVector, Tj<:Real}
```
willtebbutt, line 2:
```suggestion
    VFE(z::AbstractVector; jitter=default_jitter)
```
Might be an idea to also discuss the `jitter` term in the docstring.

willtebbutt (commented):
Broadly looks good -- just a couple of comments.

Unfortunately I think this is going to have to be a breaking change because one can no longer construct `VFE` without any arguments. I guess it's time for 0.4.

willtebbutt:
We can add `logpdf(approximation, fx, y)` and `rand(rng, approximation, fx)` at a later date, because they won't break the API.

> (n.b. dropping the approx_ prefix - should we keep it?)

I'm happy with this. If we find that it's confusing / causes problems for some reasons, we can always revert at a later date.

willtebbutt (approved):
I'm happy with this now.

Unless @st-- has any objections, I propose that we bump the version of 0.4.0-DEV, so that we can remove deprecations / make any other breaking changes we need to make over the next couple of days, and aim to release 0.4.0 on Thursday.

st--:
Ohhh I still want to rename a bunch of things... not sure we'll manage that by Thursday though!

docs/src/concrete_features.md
```diff
@@ -87,12 +86,12 @@ p_p_fx = posterior(p_fx(x[4:10], 0.1), y[4:10])
 ```julia
 Z1 = rand(rng, 4)
 Z2 = rand(rng, 3)
-p_fx = approx_posterior(VFE(), f(x[1:7], 0.1), y[1:7], f(Z))
-u_p_fx = update_approx_posterior(p_fx1, f(x[8:10], 0.1), y[8:10])
+Z = vcat(Z1, Z2)
```
st--, line 89: good catch! should we not run code snippets as some kind of doctests to ensure they at least _run_?

src/posterior_gp/approx_posterior_gp.jl
```diff
@@ -30,25 +70,25 @@ function approx_posterior(::VFE, fx::FiniteGP, y::AbstractVector{<:Real}, u::Fin
...
-    function update_approx_posterior(
-        f_post_approx::ApproxPosteriorGP,
+    function update_posterior(
+        f_post_approx::ApproxPosteriorGP{<:VFE},
         fx::FiniteGP,
         y::AbstractVector{<:Real}
     )
 
 Update the `ApproxPosteriorGP` given a new set of observations. Here, we retain the same 
 of pseudo-points.
```
st--, line 85:
```suggestion
set of pseudo-points.
```
or
```suggestion
pseudo-points.
```

src/posterior_gp/approx_posterior_gp.jl
```diff
@@ -71,30 +111,34 @@ function update_approx_posterior(
     x = vcat(f_post_approx.data.x, fx.x)
 
     cache = (m_ε=m_ε, Λ_ε=Λ_ε, U=U, α=α, z=z, b_y=b_y, B_εf=B_εf, x=x, Σy=Σy)
-    return ApproxPosteriorGP(VFE(), fx.f, cache)
+    return ApproxPosteriorGP(f_post_approx.approx, fx.f, cache)
 end
 
 """
     function update_approx_posterior(
```
st--, line 118:
```suggestion
    function update_posterior(
```

src/posterior_gp/approx_posterior_gp.jl
```diff
@@ -107,54 +151,140 @@ function update_approx_posterior(f_post_approx::ApproxPosteriorGP, u::FiniteGP)
...
+julia> isapprox(dtc(v, f(x, 0.1), y), logpdf(f(x, 0.1), y); atol=1e-6, rtol=1e-6)
+true
+```
+
+[1] - M. Seeger, C. K. I. Williams and N. D. Lawrence. "Fast Forward Selection to Speed Up
+Sparse Gaussian Process Regression". In: Proceedings of the Ninth International Workshop on
+Artificial Intelligence and Statistics. 2003
+"""
+function dtc(v::VFE, fx::FiniteGP, y::AbstractVector{<:Real})
+    return first(_compute_intermediates(fx, y, fx.f(v.z, v.jitter)))
```
st--, line 267: How about the more explicit
```suggestion
    _dtc, _ = _compute_intermediates(fx, y, fx.f(v.z, v.jitter))
    return _dtc
```
?

src/posterior_gp/approx_posterior_gp.jl
```diff
@@ -107,54 +151,140 @@ function update_approx_posterior(f_post_approx::ApproxPosteriorGP, u::FiniteGP)
...
+
+julia> elbo(v, f(x, 0.1), y) < logpdf(f(x, 0.1), y)
+true
+```
+
+[1] - M. K. Titsias. "Variational learning of inducing variables in sparse Gaussian
+processes". In: Proceedings of the Twelfth International Conference on Artificial
+Intelligence and Statistics. 2009.
+"""
+function elbo(v::VFE, fx::FiniteGP, y::AbstractVector{<:Real})
```
st--, line 235: I really like `elbo` depending on a specific type as it's quite a general concept and this makes it easy to extend it to other approximations:)

src/posterior_gp/approx_posterior_gp.jl
```diff
@@ -107,54 +151,140 @@ function update_approx_posterior(f_post_approx::ApproxPosteriorGP, u::FiniteGP)
...
+function _compute_intermediates(fx::FiniteGP, y::AbstractVector{<:Real}, fz::FiniteGP)
+    consistency_check(fx, y)
+    chol_Σy = _cholesky(fx.Σy)
+
+    A = cholesky(_symmetric(cov(fz))).U' \ (chol_Σy.U' \ cov(fx, fz))'
+    Λ_ε = cholesky(Symmetric(A * A' + I))
+    δ = chol_Σy.U' \ (y - mean(fx))
+
+    tmp = logdet(chol_Σy) + logdet(Λ_ε) + sum(abs2, δ) - sum(abs2, Λ_ε.U' \ (A * δ))
+    _dtc = -(length(y) * typeof(tmp)(log(2π)) + tmp) / 2
```
st--, line 280: Is the cast to `typeof(tmp)` necessary?
Also, is there a reason to use `log(2π)` here instead of the `log2π` constant? (that was already providing us a bit of grief with StatsFuns...)

src/posterior_gp/approx_posterior_gp.jl
```diff
@@ -107,54 +151,140 @@ function update_approx_posterior(f_post_approx::ApproxPosteriorGP, u::FiniteGP)
...
+    return _dtc, A
+end
+
+function consistency_check(fx, y)
+    @assert length(fx) == size(y, 1)
```
st--, line 285: `y` is an AbstractVector so how about
```suggestion
    @assert length(fx) == length(y)
```
?

st-- (approved):
I'm generally happy with this, I think it's a good step forward. I've left a few minor suggestions for clean-up.

One larger question I have (but I'd be happy to discuss this separately/change it in another PR if you think that makes more sense): why `posterior(VFE(z), fx, y)` and not, say, `posterior(VFE(f(z)), x, y)`? I'm mainly wondering if you had thought this through and if so what conclusions you came to, or if you hadn't thought this through what pros/cons either way you might see:)

src/posterior_gp/approx_posterior_gp.jl
st--: Ah I might've confused this with the KernelFunctions repo. But the question still stands (probably at @willtebbutt) and depending on the answer, maybe we should depend on IrrationalConstants here as well then?

rossviljoen:
Cheers for the review :)

The reason I did it that way is because you need to somehow pass both the observation noise for `x` and the jitter for `z`, so it seemed a bit cleaner to me to do 
```posterior(VFE(z; jitter=1e-6), f(x, 0.1), y)```
vs something like
```posterior(VFE(f(z, 1e-6)), x, y, 0.1)```.

Also, it's more consistent with the exact `posterior(fx, y)` in that the meaning of `fx` is the same in both.

File: docs/src/concrete_features.md (hunk 3 above)
Comment (rossviljoen) [reply docs/src/concrete_features.md:89]: good idea - could just run them as is and hide the output?

**Comment by st--** [comment]
For non-conjugate inference, would we then have something like
`posterior(VFE(z; jitter=1e-6), f(x, likelihood), y)`? :thinking:

**Comment by rossviljoen** [comment]
Yes, I suppose it would currently look like:
```
f = LatentGP(GP(kernel), likelihood, jitter)
posterior(VFE(z; jitter=1e-6), f(x), y)
```

File: src/posterior_gp/approx_posterior_gp.jl (hunk 8 above)
Comment (willtebbutt) [reply src/posterior_gp/approx_posterior_gp.jl:281]: Happy for this to change to use IrrationalConstants -- the reason to cast it was to ensure that e.g. `Float32`s don't accidentally get promoted to `Float64`s, so as long as we maintain that functionality, I don't particularly mind. (I'm not sure we actually test that that works though...)

**Comment by st--** [comment]
> ~Yes, I suppose it would currently look like:~
> 
> ~f = LatentGP(GP(kernel), likelihood, jitter)~
> ~posterior(VFE(z; jitter=1e-6), f(x), y)~
> 
> Edit: Sorry, that's nonsense

Awww, that's a shame; I thought that looked really good actually :smile:

File: src/posterior_gp/approx_posterior_gp.jl (hunk 8 above)
Comment (st--) [reply src/posterior_gp/approx_posterior_gp.jl:281]: Yeah that makes sense that you don't want to accidentally promote due to internal constants. I think
```suggestion
    _dtc = -(length(y) * log2π) + tmp) / 2
```
would be rather more pleasant to read, and KernelFunctions already depends on IrrationalConstants so it's not adding overhead:)

st--:
As discussed in person, let's just release this as 0.4.0. Ideally, we would wait with that until https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/pull/190 has been merged & released as 0.3.9. (This is blocked, in turn, by https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/344 and https://github.com/JuliaDiff/ChainRules.jl/pull/496)

src/posterior_gp/approx_posterior_gp.jl
rossviljoen: changed it to use IrrationalConstants - I think you still need the cast though because `length(y) * log2π isa Float64` unless there's some way around that?

rossviljoen:
Looking at this again, there is a problem with using `posterior(VFE(z), fx, y)` instead of `posterior(VFE(f(z)), x, y)`. For the SVGP approximation, this becomes: `posterior(SVGP(z, q))` since the data are not actually needed to construct the posterior. Therefore, the process `f` would need to be somehow passed as well.

I'm actually somewhat leaning towards reverting to the original version of:
`posterior(VFE(f(z, jitter)), f(x, σ²), y)` and just asserting that `fz.f == fx.f`? The SVGP version then becomes `posterior(SVGP(f(z, jitter), q))`. I think this gives the most natural way of passing the observation/jitter noise - it's just a shame about the redundancy of passing `f` twice...

willtebbutt:
Hmmm, yes, good point.

Thinking about this, I feel like we might be missing something from the API, but I'm not entirely sure _what_ .

I _think_ that, once an approximation has been built and an `ApproxPosteriorGP` produced, it's clear what the API should look like (just the usual AbstractGPs / FiniteGPs API).

The question, as I understand it, is
1. what the API should be to construct such an approximate posterior, and
2. what the API should be for approximate log marginal likelihood computaitons.

The only issue I have with your proposal is that you wind up with a `posterior` function that might take three arguments, or might take 1, depending upon the type of approximation being employed. I would prefer to have exactly the same API for all approximations if possible (so that one can just drop in one for the other anywhere), but maybe that's not possible in a way that seems reasonable?

For example, maybe we should be making the `VFE` / `SVGP` constructors something like
```julia
VFE(z, f(x), y; jitter)
SVGP(f(z, jitter), q)
```
? But then what do we replace `elbo`with for the `VFE` approximation? Or would the idea be that
```julia
logpdf(VFE(f(z), f(x), y), f(x), y)
logpdf(SVGP(f(z), q), f(x), y; kwargs...)
```
both yield the appropriate elbo?

Sorry, that's a bit of a brain dump. Do you have any more thoughts about the 1 vs 3 arg problem @rossviljoen ? Or perhaps I'm just being overly fussy, and we shouldn't worry about it for now?

rossviljoen:
I think that summarises the problem perfectly.

I think that your proposed constructors make the most conceptual sense, since `VFE` is essentially a special case of `SVGP` where `q` is determined completely by `fx, y`, which leads naturally to:
```
VFE(fz, fx, y)
SVGP(fz, q)
```
but then the `elbo/logpdf` has a huge amount of redundancy as you point out.

I also agree that the entire point of this PR was to get a consistent API across all approximations. To that end, I'm in favour of the API looking like:
```
posterior(approx, fx, y)
elbo(approx, fx, y)
```
because that is also the most consistent with the non-approx API.
Then, we'd have
```
approx = VFE(f(z, jitter))
approx = SVGP(f(z, jitter), q)
```
and simply define
```
posterior(SVGP(fz, q), fx, y) = posterior(SVGP(fz, q))
```
?

You're right though, it still doesn't quite feel 100%

willtebbutt:
At this point in time, I think it's more important to having _something_ in, so that we can crack on with building the other functionality, even if we refactor later when we realise there's a better way to do stuff.

> but then the elbo/logpdf has a huge amount of redundancy as you point out.

I suppse that one could just delay computing stuff with `VFE` until either `posterior` or `logpdf` / `elbo` is called?

> and simply define

Would it not simply be `posterior(approx)` in both cases, since `VFE` would now have enough information to know how to compute the approximate posterior without any additional arguments?

rossviljoen:
Ok, I've changed it again to use the style

```
posterior(VFE(f(z)), f(x), y)
elbo(VFE(f(z)), f(x), y)
```

as well as added assertions that fz.f == fx.f

src/sparse_approximations.jl
```diff
@@ -8,20 +22,39 @@ struct ApproxPosteriorGP{Tapprox,Tprior,Tdata} <: AbstractGP
...
+true
+```
 
 [1] - M. K. Titsias. "Variational learning of inducing variables in sparse Gaussian
 processes". In: Proceedings of the Twelfth International Conference on Artificial
 Intelligence and Statistics. 2009.
 """
-function approx_posterior(::VFE, fx::FiniteGP, y::AbstractVector{<:Real}, u::FiniteGP)
+function posterior(vfe::VFE, fx::FiniteGP, y::AbstractVector{<:Real})
+    @assert vfe.fz.f == fx.f
```
willtebbutt, line 52:
```suggestion
    @assert vfe.fz.f === fx.f
```
Correct me if I'm wrong on this, but I suspect that we want to ensure that it's _literally_ the same object.

willtebbutt (approved):
I'm happy for this to go in, once a minor version bump has been added and my comment addressed.

PR #196, opened 2021-08-12
clean up directory structure

st--:


yebai:
>  would you consider keeping gp.jl rather than calling it gp_prior.jl.

I also like `gp.jl` a bit better than `gp_prior.jl`.

PR #203, opened 2021-08-15
Bump CRC

willtebbutt:
Supercedes #190 and #191 .


willtebbutt:
No idea whatsoever what's going on with the docs.

willtebbutt:
For some reason the docs CI isn't able to find the new version of KernelFunctions -- not sure why that's happening. Will try again later.

devmotion:
Maybe there's some problem with the package server right now and it serves an outdated registry? You could try to set  the environment variable `JULIA_PKG_SERVER` to an empty string, then it should download the registry with git.

willtebbutt:
Looking promising. @devmotion unless you can think of anything else, I think this is probably good to go.

devmotion (approved):
LGTM, we can't do much about the failing PPL tests :shrug:

willtebbutt:
Yeah. 🤞 they'll start working again at some point in the future without any intervention from us.

PR #207, opened 2021-08-16
Tweak API Docs

willtebbutt:
Just some minor alterations. Nothing serious, just stuff I thought could be expressed better while I was re-reading the docs.


yebai (approved):
Looks good to me.

PR #210, opened 2021-08-20
Doc plotting (continuation of #174)

devmotion:
This PR is a continuation of https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/pull/174. I updated it with the latest changes in the master branch and added docstrings in a similar way as done for `@userplot` automatically: the same docstring is added to `plot(args...; kwargs...)`, `plot!(args...; kwargs...)` and `plot!(::AbstractPlot, args...; kwargs...)`.

I wonder if the documentation about plotting on the website would be easier to find if it would be moved to a separate page. So far I only moved it to the top of the file.

I guess one might want to consider modifying the `@recipe` macro in RecipesBase such that docstrings for `@recipe` are added to these plot calls automatically (similar to `@userplot`).


willtebbutt (approved):
Happy for this to go in as-is.

The only suggestion that I would make would be to also set the ribbon_scale to 2 in the sparse and exact inference plots, just for the sake of consistency.

theogf:
I think the separation of the sections in "Concrete Features" could be made more clear, right now "Plotting" vs "Setup" is a bit confusing...

devmotion:
IMO this page is not very helpful and mainly a list of commands without explanations. Since this is not related to the plotting and a more general issue, I think this should be addressed separately though. I moved the plotting docs to the top of the page so they are more prominently placed and easier to discover. Maybe it would be cleaner to move them to a separate page as suggested above?

theogf:
Let's leave it here for now and consider another structure in a different issue

willtebbutt:
I'm happy for this to be merged as-is, or for the plotting docs to be made into their own section (since they're really very important). Either way, my approval stands.

PR #213, opened 2021-08-24
Plotting updates

st--:


st--:
> If you specify samples = n, layout = n you might actually want to have a label for each series/sample which would end up in different subplots. The example might be a bit constructed but it shows that this might not always be desired.

I would argue that even in the samples=layout=n case it might be cleaner to have just one legend. But if you do want multiple labels, it's as easy as passing `label=fill("samples", (1, n))`, whereas the (in my opinion much more frequent use-case) of plotting ~100 samples and only wanting a single label seems rather clunkier...

st--:
@devmotion I've added an explanation to the docstring, hopefully that makes things sufficiently clear for users:) I found the previous behaviour of `sampleplot(x, f; samples=100, label="samples")` deeply confusing... (as a user, I pass in a _single_ `f`, so I would expect to be able to pass in a single `label` to get a single legend entry)

st--:
I've now also added a use-case in the regression 1d example: )

examples/regression-1d/script.jl
```diff
@@ -212,19 +212,18 @@ mean(logpdf(gp_posterior(x_train, y_train, p)(x_test), y_test) for p in samples)
...
+        0:0.02:1,
+        gp_posterior(x_train, y_train, p);
+        samples=5,
+        label=(i == 1 ? "samples" : nothing),
+    )
 end
-scatter!(plt, x_test, y_test; label="Test Data")
+scatter!(plt, x_train, y_train; label="Train Data", markercolor=1)
+scatter!(plt, x_test, y_test; label="Test Data", markercolor=2)
 plt
```
st--, line 227: NB- is the explicit passing around of `plt` actually needed (/ recommended)? It would seem to me it should work exactly the same without it (the last `scatter!` also returns the plot object).

examples/regression-1d/script.jl
```diff
@@ -212,19 +212,18 @@ mean(logpdf(gp_posterior(x_train, y_train, p)(x_test), y_test) for p in samples)
...
+        plt,
+        0:0.02:1,
+        gp_posterior(x_train, y_train, p);
+        samples=5,
+        label=(i == 1 ? "samples" : nothing),
+    )
 end
-scatter!(plt, x_test, y_test; label="Test Data")
+scatter!(plt, x_train, y_train; label="Train Data", markercolor=1)
+scatter!(plt, x_test, y_test; label="Test Data", markercolor=2)
```
st--, line 226: It seems like the only way of getting the scatter markers of train/test data _on top of_ the lines for the samples is by having the call later. That then also affects the order of legend entries. I wish there was something straightforward like the z-order in matplotlib! If any of you has any ideas for how to get that behaviour, I'd love to find out (google didn't really help me here)..

devmotion:
Related: https://discourse.julialang.org/t/plots-only-one-series-as-primary-plot/24616/8

Apparently, the plots maintainers think it would be more natural to concatenate all data in a single vector separated with `NaN` :grimacing: Maybe we could add a keyword argument `uniquelabels=true`, as suggested in the discussion? I just think it is a bit unfortunate if we completely rule out the possibility to provide a string that should be duplicated for each plot as it is the current behaviour (BTW I have never managed to get a joint legend with subplots but maybe I didn't try long enough - the docs are quite difficult to navigate :smile: ). Since probably most users would want only a single label we could just set `uniquelabels=true`.

examples/regression-1d/script.jl
devmotion: No, I think you just have to plot in whatever order you want them to appear and then you have to live with the legend. At least I don't know of any other way.

examples/regression-1d/script.jl
devmotion: Not sure :shrug: I know that I had to use it sometimes since the plot was not returned from `scatter!` (or some other inplace function) but maybe it was fixed.

src/util/plotting.jl
```diff
@@ -82,16 +82,18 @@ Plot samples from the projection `f` of a Gaussian process versus `x`.
     Make sure to load [Plots.jl](https://github.com/JuliaPlots/Plots.jl) before you use
     this function.
 
+This recipe handles the `label` argument in a custom way: when plotting multiple samples and `label` is a string, only a single element is added to the legend. To give each sample the same legend entry, pass an explicit column matrix, e.g. `label=fill("sample", (1, samples))`.
```
devmotion, line 85: Can we stick to the 92 characters limit?

src/util/plotting.jl
```diff
@@ -115,18 +117,20 @@ SamplePlot((f,)::Tuple{<:FiniteGP}) = SamplePlot((f.x, f))
...
 @recipe function f(sp::SamplePlot)
-    nsamples::Int = get(plotattributes, :samples, 1)
+    nsamples::Int = pop!(plotattributes, :samples, 1)
     samples = rand(sp.f, nsamples)
 
+    col_label = pop!(plotattributes, :label, "")
+    if col_label isa String
+        # blank labels for all but first column, e.g. ["samples" "" "" ...]
+        col_label = [col_label fill("", nsamples - 1)...]
+    end
```
devmotion, line 127: It is better to avoid splatting and to use `hcat`. Also sometimes the compiler becomes confused if one reassigns a variable with a different type.
```suggestion
    _label = pop!(plotattributes, :label, "")
    col_label = if _label isa String
        # blank labels for all but first column, e.g. ["samples" "" "" ...]
        hcat(_label, fill("", nsamples - 1))
    else
        _label
    end
```

st--:
> I just think it is a bit unfortunate if we completely rule out the possibility to provide a string that should be duplicated for each plot as it is the current behaviour (BTW I have never managed to get a joint legend with subplots but maybe I didn't try long enough - the docs are quite difficult to navigate smile ). Since probably most users would want only a single label we could just set `uniquelabels=true`.

It's not impossible: does the explanation I added to the docstring, using `label=fill("single label", (1, samples))`, not satisfy you for that (rare?) use-case?

devmotion:
> It's not impossible: does the explanation I added to the docstring, using `label=fill("single label", (1, samples))`, not satisfy you for that (rare?) use-case?

What I meant more precisely: it is not possible anymore if you pass a string as label even though in the Plots ecosystem usually that would give you the same label for each series. And I think `label=fill("single label", (1, samples))` is more cumbersome than `label="single label", uniquelabels=false` since you have to use `fill` and pass the correct number of labels.

src/util/plotting.jl
st--: > It is better to avoid splatting and to use hcat.

For my understanding: What is better about that ?

(That's actually what I had first but then I thought you'd come and tell me that the shorter splatting would be more idiomatic so then changed it to that :joy:)

src/util/plotting.jl
st--: woop, meant to do `Vgq` and forgot...

src/util/plotting.jl
devmotion: Splatting puts more stress on the compiler. It's fine if you use e.g. small tuples but especially when you work with possibly longer array it is good to avoid it if you can.

st--:
> in the Plots ecosystem usually that would give you the same label for each series. 

Yes and that's what I'd expect if I pass in a series of `f`s! As a user, samples>1 doesn't give me any indication that it's suddenly multiple series behind the scenes (it might've just been samples connected with NaNs as you mentioned). But if it's gonna make you much happier I can add the `uniquelabels`.

devmotion:
My opinion is just:
- If we treat the samples as multiple series, then we should not deviate too much from what is the standard for multiple series in the Plots.jl ecosystem. I think a `uniquelabels` keyword argument would be the most straightforward solution since it would still allow to supply a string as label but get a legend with multiple labels without having to construct the list of labels manually.
- If we treat the samples as a single series, then we should concatenate them in a vector with the different samples separated by `NaN`. This is the only way to ensure that downstream transformations in the recipe pipeline treat the data as a single series. The label issue would be resolved automatically in this case.

I am leaning more and more towards the second option.

st--:
> I am leaning more and more towards the second option.

Sounds good :+1:  I think that makes most sense conceptually as well. I'll have a go at that:)

src/util/plotting.jl
```diff
@@ -115,18 +118,16 @@ SamplePlot((f,)::Tuple{<:FiniteGP}) = SamplePlot((f.x, f))
 SamplePlot((x, gp)::Tuple{<:AbstractVector,<:AbstractGP}) = SamplePlot((gp(x, 1e-9),))
 
 @recipe function f(sp::SamplePlot)
-    nsamples::Int = get(plotattributes, :samples, 1)
+    nsamples::Int = pop!(plotattributes, :samples, 1)
     samples = rand(sp.f, nsamples)
 
+    flat_x = vcat(Iterators.flatten(zip(Iterators.repeated(sp.x), fill(NaN, nsamples)))...)
+    flat_f = vcat(Iterators.flatten(zip(eachcol(samples), Iterators.repeated([NaN])))...)
```
st--, line 125: @devmotion happy for you to suggest a better way - this is what I could come up with so far ;)

st--: Apparently the splatting is bad? but I can't figure out a way to get around it

devmotion: My suggestion would be
```suggestion
    flat_x = repeat(vcat(sp.x, NaN), nsamples)
    flat_f = vec(vcat(samples, fill(NaN, 1, nsamples)))
```

st--: ah, I literally just came upon the same solution myself right before looking at your comment :tada:

st--: Except the "interleave with NaN" doesn't actually seem to work, as `seriestype = :line` makes it sort by x :(

So we could remove the `seriestype --> :line` below, but then you must pass in an ordered `x` for the `sampleplot` to make any sense. Would you be happy with _that_?

src/util/plotting.jl
```diff
@@ -115,18 +118,17 @@ SamplePlot((f,)::Tuple{<:FiniteGP}) = SamplePlot((f.x, f))
 SamplePlot((x, gp)::Tuple{<:AbstractVector,<:AbstractGP}) = SamplePlot((gp(x, 1e-9),))
 
 @recipe function f(sp::SamplePlot)
-    nsamples::Int = get(plotattributes, :samples, 1)
+    nsamples::Int = pop!(plotattributes, :samples, 1)
     samples = rand(sp.f, nsamples)
 
+    flat_x = repeat(vcat(sp.x, NaN), nsamples)
+    flat_f = vcat(samples, fill(NaN, 1, nsamples)) |> vec
```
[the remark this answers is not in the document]
st--: I found the `|>` version more intuitively readable, but okay... can't argue with a bot :stuck_out_tongue_closed_eyes:

src/util/plotting.jl
devmotion: We could also just sort `x`, couldn't we?

st--:
I think it makes more sense to treat the multiple samples as a single series, but I also don't think it needs to be `sampleplot`'s responsibility to sort the x values. But I'm not sure what to do about the failing plotting tests - they seem to test some very specific implementation details (which doesn't strike me as particularly valuable tests); we could change them so we just test that the length of the "new" x is equal to samples * (length(x) + 1) ? or delete them? what do you prefer?

devmotion:
I'm happy if the tests are improved. I think in the tests we should check the plotting attributes that `sampleplot` defines. So I think it makes sense to try to check that `x` is correct, `y` looks reasonable, and the default values for other attributes correspond to what we define in `sampleplot`. As long as Plots does the correct thing even if the `x` values are not sorted, I think we should not sort them.

examples/regression-1d/script.jl
```diff
@@ -212,19 +212,18 @@ mean(logpdf(gp_posterior(x_train, y_train, p)(x_test), y_test) for p in samples)
...
-for p in samples[(end - 100):end]
-    sampleplot!(plt, 0:0.02:1, gp_posterior(x_train, y_train, p); samples=5)
+plt = plot(; xlim=(0, 1), xlabel="x", ylabel="y", title="posterior (AdvancedHMC)")
+for (i, p) in enumerate(samples[(end - 100):end])
+    sampleplot!(
+        plt,
+        0:0.02:1,
+        gp_posterior(x_train, y_train, p);
+        samples=5,
+        label=(i == 1 ? "samples" : nothing),
```
devmotion, line 222: Maybe just
```suggestion
        label=(i == 1 ? "samples" : ""),
```
? I did not know that `label=nothing` is a thing :smile:

examples/regression-1d/script.jl
```diff
@@ -212,19 +212,18 @@ mean(logpdf(gp_posterior(x_train, y_train, p)(x_test), y_test) for p in samples)
...
+    )
 end
-scatter!(plt, x_test, y_test; label="Test Data")
+scatter!(plt, x_train, y_train; label="Train Data", markercolor=1)
```
devmotion, line 225: I guess the colors are not consistent with the other plots if we don't use `markercolor`?

src/util/plotting.jl
```diff
@@ -115,18 +120,16 @@ SamplePlot((f,)::Tuple{<:FiniteGP}) = SamplePlot((f.x, f))
...
     # Set default attributes
-    seriestype --> :line
-    linealpha --> 0.2
-    markershape --> :circle
-    markerstrokewidth --> 0.0
-    markersize --> 0.5
-    markeralpha --> 0.3
+    linealpha --> 0.35
     seriescolor --> "red"
     label --> ""
```
devmotion, line 132: Do we need/want these defaults?

devmotion (commented):
Looks good, I just had some minor questions :slightly_smiling_face:

examples/regression-1d/script.jl
st--: they weren't consistent before; I've now made them consistent (check the deployed docs)

examples/regression-1d/script.jl
st--: Yeah, one of the [complaints about Plots.jl](http://juliaplots.org/MakieGallery.jl/stable/why-makie.html) is that there are plenty of different ways of saying the same thing!
I find `label = nothing` to be a bit more self-explanatory, but feel free to add&commit your suggestion if you feel moderately strongly about it:)

src/util/plotting.jl
st--: Yeah, I really like this default for `label` and would keep it. Otherwise it says `y1` which is worse. If you want to get rid of the red for `seriescolor`, we can remove it here as well, but I'm also happy to leave it as is for now.

devmotion: Yes, the default labels aren't great. Maybe let's just remove the series color? I don't think red is a particularly great choice and we could just let Plots pick one, according to whatever theme is used.

examples/regression-1d/script.jl
devmotion: No, I don't feel strongly about it. The main reason would be to avoid having two different types here but I'm fine with `nothing` as well.

PR #218, opened 2021-08-27
RFC: Sketch of vector-valued GP functionality

willtebbutt:
**Summary**

What: Proposal for a vector-valued GP.

Why: It might be a convenient thing to have around for some use-cases in which people really do think about their multi-output GPs as being vector-valued. 

Why: it's easy to provide, because we already have the functionality, we just need to package it nicely.

**Proposed changes**

Add a vector-valued GP. Not proposing to export at this point in time, because I don't know what I even think about it.

Contains a sketch of the implementation in which the intent is hopefully clear. Does not contain proper testing, and I don't believe that the API is complete.

I considered not bothering with this, and I still think that's a good option -- I really don't know whether or not I like this, but I can imagine that it would ease cognitive burden in some common workflows.

**Breaking changes**

No breaking changes.

### DIFF
--- a/src/vector_valued_gp.jl
+++ b/src/vector_valued_gp.jl
@@ -73,3 +73,20 @@ function Distributions.logpdf(vx::FiniteVectorValuedGP, Y::AbstractMatrix{<:Real
     # Compute logpdf using FiniteGP.
     return logpdf(fx, y)
 end
+
+function posterior(vx::FiniteVectorValuedGP, Y::AbstractMatrix{<:Real})
+
+    # Construct equivalent FiniteGP.
+    x_f = KernelFunctions.MOInputIsotopicByOutputs(vx.x, vx.v.num_outputs)
+    f = vx.v.f
+    fx = f(x_f, vx.Σy)
+
+    # Construct flattened-version of observations.
+    y = vec(Y)
+
+    # Construct posterior AbstractGP
+    f_post = posterior(fx, y)
+
+    # Construct a new vector-valued GP.
+    return VectorValuedGP(f_post, vx.v.num_outputs)
+end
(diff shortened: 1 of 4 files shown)


src/vector_valued_gp.jl
```diff
@@ -0,0 +1,92 @@
...
+end
+
+(f::VectorValuedGP)(x...) = FiniteVectorValuedGP(f, x...)
+
+function Statistics.mean(vx::FiniteVectorValuedGP)
+
+    # Construct equivalent FiniteGP.
+    x_f = KernelFunctions.MOInputIsotopicByOutputs(vx.x, vx.v.num_outputs)
+    f = vx.v.f
+    fx = f(x_f, vx.Σy)
```
st--, line 23: The amount of code-duplication in here seems to beg for something like
```julia
function _vector_equivalent_finitegp(vx::FiniteVectorValuedGP)
    x_f = KernelFunctions.MOInputIsotopicByOutputs(vx.x, vx.v.num_outputs)
    f = vx.v.f
    fx = f(x_f, vx.Σy)
    return fx
end
```
(or `_construct_equivalent_finitegp`)
what do you think?

Could also have a
```julia
function _vectorvaluedgp_reshape(vx::FiniteVectorValuedGP, a)
    return reshape(a, length(vx.x), vx.v.num_outputs)
end
```
NB: should the return type be a ColVecs ?

src/vector_valued_gp.jl
```diff
@@ -0,0 +1,92 @@
...
+
+    # Compute quantity under equivalent FiniteGP.
+    m = mean(fx)
+
+    # Construct the matrix-version of the quantity.
+    M = reshape(m, length(vx.x), vx.v.num_outputs)
+    return M
```
st--, line 30: Could also have a
```julia
function _vectorvaluedgp_reshape(vx::FiniteVectorValuedGP, a)
    return reshape(a, length(vx.x), vx.v.num_outputs)
end
```
NB: should the return type be a ColVecs ?

PR #225, opened 2021-09-29
define Base.length(::LatentFiniteGP)

st--:
Just as there is `length(::FiniteGP)` - means we have to rely less on implementation details (the `fx` field) e.g. for dimension consistency checking.:)


willtebbutt:
Lovely. Could you add a 1-liner test just to ensure that this functionality gets run once in our test suite? Other than that, just needs a patch bump.

willtebbutt (approved):
Perfect. Thanks!

PR #239, opened 2021-11-04
remove superfluous _symmetric() calls

st--:
**Summary**
Remove `_symmetric()` calls on user-provided matrices that should be pos-def (and hence symmetric) already, as discussed in https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/pull/236#discussion_r738464168

**Proposed changes**
We use `_symmetric()` to ensure we can compute cholesky of matrices that _should_ be posdef (mathematically) but might not be (numerically, due to small rounding errors in matmul & co). This is the right thing to do for intermediate matrices, but when applied to user input directly, this risks just hiding mistakes. I.e., if a user provides an arbitrary (non-pos-def) matrix, it would simply mask the mistake, happily compute an answer (which will be misleading), instead of crashing & giving the user an indication that there's a mistake somewhere.

We could add a check to `_symmetric` that the return value is approximately the same as the input. We could also explicitly check that the user input is symmetric (or nearly so), or enforce it to be some kind of PDMat type.

**Breaking changes**
I would consider this a bugfix, the only thing that would break is already-broken usage.

### DIFF
--- a/src/sparse_approximations.jl
+++ b/src/sparse_approximations.jl
@@ -51,7 +51,7 @@ Intelligence and Statistics. 2009.
 function posterior(vfe::VFE, fx::FiniteGP, y::AbstractVector{<:Real})
     @assert vfe.fz.f === fx.f
 
-    U_y = _cholesky(_symmetric(fx.Σy)).U
+    U_y = _cholesky(fx.Σy).U
     U = cholesky(_symmetric(cov(vfe.fz))).U
 
     B_εf = U' \ (U_y' \ cov(fx, vfe.fz))'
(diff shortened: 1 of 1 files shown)


devmotion:
Thanks for opening this PR!

> I would consider this a bugfix, the only thing that would break is already-broken usage.

I disagree: It breaks working code that operates with matrices that mathematically are guaranteed to be symmetric and positive definite but are not due to numerical inaccuracies and floating point issues.


It is also a bit unfortunate that such numerical issues are not caught when `FiniteGP` is constructed. One alternative approach that is eg used in Distances and possibly soon in PDMats is to check positive semi-definiteness at construction and then assume internally that the matrix is positive semi-definite and eg wrap it in `Symmetric` if helpful for downstream methods:
```julia
struct MyType{T}
    A::T

    function MyType{T}(A::T; check_posdef::Bool=true) where {T}
         if check_posdef
             eigmin(A) >= 0 || error("A is not positive semi-definite")
         end
         return new{T}(A)
    end
end
```
`eigmin` is fast for `ScalMat`, `Diagonal` etc. and the check could also be disabled if desired. And internally we could use numerical optimizations such as `_symmetric` without making incorrect assumptions.

willtebbutt:
This PR has gone stale, so I'm going to close. @st-- please do re-open if you want to push it forwards.

PR #240, opened 2021-11-05
Add Mauna Loa example notebook

st--:
Now that it works:)


examples/mauna-loa/script.jl
```diff
@@ -0,0 +1,169 @@
...
+          end
+    end
+
+    result = optimize(
+        Optim.only_fg!(fg!),
+        θ_flat_init,
+        optimizer,
+        options;
+        inplace=false,
+    )
```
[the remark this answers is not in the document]
theogf: That would be great if the formatter would stop doing this...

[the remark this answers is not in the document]
st--: as in, you preferred the original list?

examples/mauna-loa/CO2_data.csv
```diff
@@ -0,0 +1,682 @@
+1958.2083333333333,315.71
```
devmotion, line 1: Can we just download the dataset in the script? We could even use https://github.com/oxinabox/DataDeps.jl but maybe that's overkill here.

examples/mauna-loa/script.jl
```diff
@@ -0,0 +1,170 @@
...
+    data = CSV.read("CO2_data.csv", Tables.matrix; header=0)
+    year = data[:, 1]
+    co2 = data[:, 2]
+
+    #Split the data into training and testing data
+    xtrain = year[year .< 2004]
+    ytrain = co2[year .< 2004]
+    xtest = year[year .>= 2004]
+    ytest = co2[year .>= 2004]
+end;
```
devmotion, line 21: `begin` doesn't introduce a new scope (which would also be undesired here):
```suggestion
data = CSV.read("CO2_data.csv", Tables.matrix; header=0)
year = data[:, 1]
co2 = data[:, 2]

#Split the data into training and testing data
xtrain = year[year .< 2004]
ytrain = co2[year .< 2004]
xtest = year[year .>= 2004]
ytest = co2[year .>= 2004]
```

examples/mauna-loa/script.jl
```diff
@@ -0,0 +1,170 @@
...
+    plot(; xlabel="year", ylabel="CO2", legend=:bottomright)
+    scatter!(xtrain, ytrain; ms=2, label="train")
+    return scatter!(xtest, ytest; ms=2, label="test")
+end
+
+plotdata()
+
+# We will use ParameterHandling.jl for handling the (hyper)parameters of our model. We represent all required parameters as a nested NamedTuple:
+
+#! format: off
```
devmotion, line 33: Does this show up in the markdown and Jupyter output?

examples/mauna-loa/script.jl
```diff
@@ -0,0 +1,170 @@
...
+begin
+    SE(θ) = θ.σ^2 * with_lengthscale(SqExponentialKernel(), θ.ℓ)
+    # PeriodicKernel is broken, see https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/issues/389
+    #Per(θ) = θ.σ^2 * with_lengthscale(PeriodicKernel(; r=[θ.ℓ/2]), θ.p)  # NOTE- discrepancy with GaussianProcesses.jl
+    function Per(θ)
+        return θ.σ^2 * with_lengthscale(SqExponentialKernel(), θ.ℓ) ∘
+               PeriodicTransform(1 / θ.p)
+    end
+    RQ(θ) = θ.σ^2 * with_lengthscale(RationalQuadraticKernel(; α=θ.α), θ.ℓ)
+end
```
devmotion, line 72: Again I suggest removing the `begin` and `end`. Also I guess the comment could be removed?

examples/mauna-loa/script.jl
```diff
@@ -0,0 +1,170 @@
...
+# The call to `ParameterHandling.value` is required to replace the constraints (such as `positive`) with concrete numbers:
+
+fpost_init = build_posterior_gp(ParameterHandling.value(θ_init))
+
+# This is what the GP fitted to the data looks like for the initial choice of kernel hyperparameters:
+
+let
+    plotdata()
+    plot!(fpost_init(1900:0.2:2050))
+end
```
devmotion, line 100:
```suggestion
plotdata()
plot!(fpost_init(1900:0.2:2050))
```

examples/mauna-loa/script.jl
```diff
@@ -0,0 +1,170 @@
...
+
+# To improve the fit, we want to maximize the (log) marginal likelihood with respect to the hyperparameters.
+# Optim.jl expects to minimize a loss, so we define it as the negative log marginal likelihood:
+
+function loss(θ)
+    fx = build_finite_gp(θ)
+    # TODO: define logpdf(::PosteriorGP) in AbstractGPs.jl?
```
devmotion, line 107: Remove the comment?
```suggestion
```

examples/mauna-loa/script.jl
```diff
@@ -0,0 +1,170 @@
...
+    lml = logpdf(fx, ytrain)
+    return -lml
+end
```
devmotion, line 110: Maybe just
```suggestion
loss(θ) = -logpdf(build_finite_gp(θ), ytrain)
```
?

examples/mauna-loa/script.jl
```diff
@@ -0,0 +1,170 @@
...
+
+# In the future, we are planning to provide the following utility function as
+# part of JuliaGaussianProcesses -- for now, we just define it inline.
```
devmotion, line 113: Also remove (and maybe open an issue :slightly_smiling_face:)?
```suggestion
```

examples/mauna-loa/script.jl
```diff
@@ -0,0 +1,170 @@
...
+# The L-BFGS parameters were chosen because they seem to work well empirically.
+# You could also try with the defaults.
+
+default_optimizer = LBFGS(;
+    alphaguess=Optim.LineSearches.InitialStatic(; scaled=true),
+    linesearch=Optim.LineSearches.BackTracking(),
+)
```
devmotion, line 120: Maybe keep it simple for the tutorial and just use the default settings (if they lead to acceptable results)?

examples/mauna-loa/script.jl
```diff
@@ -0,0 +1,170 @@
...
+
+function optimize_loss(loss, θ_init; optimizer=default_optimizer, maxiter=1_000)
+    options = Optim.Options(; iterations=maxiter, show_trace=true)
+
+    θ_flat_init, unflatten = ParameterHandling.value_flatten(θ_init)
+    loss_packed = loss ∘ unflatten
+
+    # https://julianlsolvers.github.io/Optim.jl/stable/#user/tipsandtricks/#avoid-repeating-computations
+    function fg!(F, G, x)
+        if F != nothing && G != nothing
```
devmotion, line 130: One should basically always use `!== nothing` and `=== nothing`:
```suggestion
        if F !== nothing && G !== nothing
```
Here it seems a bit nicer though to just dispatch, i.e., to define
````julia
function fg!(::Nothing, ::Nothing, x)
...
end
function fg!(f, ::Nothing, x)
...
end
function fg!(::Nothing, G, x)
...
end
```

examples/mauna-loa/script.jl
```diff
@@ -0,0 +1,170 @@
...
+end
+
+# We now run the optimisation:
+
+θ_opt, result = optimize_loss(loss, θ_init)
+result
+
+# The final value of the log marginal likelihood is:
+
+-result.minimum
```
devmotion, line 155: I guess it doesn't matter, we could also just report the negative log marginal likelihood.

examples/mauna-loa/script.jl
```diff
@@ -0,0 +1,170 @@
...
+# This is the kernel with the point-estimated hyperparameters:
+
+fpost_opt.prior.kernel
+
+# And, finally, this is the visualization of the posterior GP:
+
+let
+    plotdata()
+    plot!(fpost_opt(1900:0.2:2050))
+end
```
devmotion, line 170:
```suggestion
plotdata()
plot!(fpost_opt(1900:0.2:2050))
```

devmotion (commented):
Great that it works now!

examples/mauna-loa/script.jl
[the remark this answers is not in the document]
theogf: Yes, it looks generally clearer

examples/mauna-loa/script.jl
theogf: I guess this is coming from Pluto :D

examples/mauna-loa/script.jl
st--: Tell that to the people of Optim.jl! I just copied it from https://julianlsolvers.github.io/Optim.jl/stable/#user/tipsandtricks/#avoid-repeating-computations ...

examples/mauna-loa/script.jl
st--: The whole function is supposed to disappear in a "make AbstractGPs nice to use" helper module, I just inlined the whole of https://github.com/JuliaGaussianProcesses/EcosystemIntegrationTesting/blob/main/api_testing/utils_optimize_loss.jl in here.

st--: Until then, I was thinking it'd be good to have this copy&pastable from the notebook.

examples/mauna-loa/CO2_data.csv
st--: I'd very much rather not. It'll fail, eventually (these things always do), and if it's just a small toy dataset it's so much nicer if it just works out of the box and doesn't require an internet uplink (beyond having already pulled the repo)...

What's your motivations for doing so ?

examples/mauna-loa/script.jl
st--: Yeah, that's left over from Pluto, but you're right in this format it's probably neater to do away with the begin-end. Just have to be careful so it doesn't end up printing all the arrays!

examples/mauna-loa/script.jl
st--: Good point. If it does, I suggest we change our literate.jl to simply filter out those two lines.

examples/mauna-loa/script.jl
st--: In this case I'd like to keep the let-end to delineate the plotting block

examples/mauna-loa/script.jl
st--: Now that I've moved it into #241, yes:)

examples/mauna-loa/CO2_data.csv
devmotion: I think it does not belong into this repo (more extremely, eg. I don't think we would want to include the MNIST dataset either) and is already publicly available. Also linking to and downloading from the original data source makes it clear to users what the authors and license of the dataset are and eg how to cite. Otherwise we would have to include this information in the repo, in my opinion.

One already has to have an internet connection for downloading the tutorial, so I don't see this as an issue. With DataDeps, the dataset will also be cached in the Julia directories automatically, so one can rerun the tutorial without any problems also without internet connection.

examples/mauna-loa/script.jl
devmotion: Isn't it simpler to just remove this line in this case? 😅

examples/mauna-loa/script.jl
devmotion: But there are no variables declared here? So the `let` block really doesn't serve any purpose. And it's already a separate code block, so it's already clearly separated from the rest. IMO the `let` is just surprising and potentially confusing for newcomers.

examples/mauna-loa/script.jl
devmotion: Unrelated from style issues, one should basically never compare `nothing` with `==` or `!=`, so let's change this at least.

examples/mauna-loa/script.jl
devmotion: It seems this comment was lost, at least in my Github interface.

examples/mauna-loa/script.jl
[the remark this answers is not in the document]
st--: @theogf do you want to take it up with BlueStyle/JuliaFormatter.jl, change the line length to 80 (which would make this line break up, but possibly others too), wrap it in `#! format: off/on`, or grumble and accept the formatter as it has spoken?

examples/mauna-loa/script.jl
st--: yes, but in this case the auto-formatted version is much harder to read

examples/mauna-loa/script.jl
st--: i'm happy to add a comment to the let block:) it took me surprisingly long until I realised that it worked really well for this purpose, and i would've liked to come across it sooner- so I see it as an advantage if newcomers come across it here

examples/mauna-loa/script.jl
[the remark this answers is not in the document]
theogf: I think I'll just grumble and mumble in my corner

examples/mauna-loa/script.jl
st--: Yeah, because I removed the comment once I had moved it into an issue (thanks for pointing it out, should've done that before moving it into this PR).

I would prefer to keep it on multiple lines, to make it clearer what's going on. Now that I'm looking at it again, I'm wondering if it'd be better to actually call the variable `log_marginal_likelihood`, or to add a comment to the line calling `logpdf` `# log marginal likelihood`...?

examples/mauna-loa/script.jl
st--: Also opened a PR upstream: https://github.com/JuliaNLSolvers/Optim.jl/pull/961

examples/mauna-loa/script.jl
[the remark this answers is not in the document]
devmotion: I'm sure that neither JuliaFormatter nor BlueStyle will move away from the Julia default of 92 characters. If we want 80 characters in the docs, the correct thing to do would be to add `.JuliaFormatter.toml` files to examples/ and docs/ that specify Blue style and an 80 characters limit.

examples/mauna-loa/script.jl
devmotion: We should just configure the formatter if we want only 80 characters in the docs and examples (see my other comment for details).

examples/mauna-loa/CO2_data.csv
devmotion: Another, IMO important, argument against including the csv file here is that it implies that one always has to copy around the data when the example is copied or moved somewhere, and one has to ensure that Julia is started with the correct path settings since otherwise the example won't run. Basically, this is the same reason for why Pluto recommends to avoid local resources. It's also illustrated by the fact that all these changes in the literate script are required (which could be reverted if the csv file is not included and the example is self contained).

docs/literate.jl
```diff
@@ -40,6 +40,10 @@ function preprocess(content)
     # remove VSCode `##` block delimiter lines
     content = replace(content, r"^##$."ms => "")
 
+    # When run through Literate, the actual @__DIR__ macro points to the OUTDIR
+    # Instead, replace it with the directory in which the script itself is located:
+    content = replace(content, r"@__DIR__" => EXAMPLEPATH)
```
devmotion, line 45: We could avoid this hack (I wonder if it breaks anything?) when the examples are self-contained and the csv file is not included. I think this would be generally preferable (see my other comments) and not only in the Literate/CI setting.

examples/mauna-loa/script.jl
st--: With 80 character lines this block would still look rather less readable than breaking it up like this. In general I'm very happy to have an auto-formatter, but in some cases it's just more readable to do it manually

examples/mauna-loa/CO2_data.csv
st--: DataDeps seems reasonable, I'd be happy for you to either leave a code suggestion or make a follow-up PR that changes it from local CSV to DataDeps handling!

examples/mauna-loa/script.jl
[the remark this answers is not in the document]
st--: The "take it up with BlueStyle/JuliaFormatter" was regarding a different heuristic for breaking up lines (e.g. when there's more than, say 3, different arguments, break it up even if it would fit within the line limit). Of course if it was just about the line length, we would change that in the .JuliaFormatter.toml, that's what I meant:)

examples/mauna-loa/script.jl
st--: I've added a replacement for these lines.

examples/mauna-loa/script.jl
```diff
@@ -0,0 +1,236 @@
...
+#md nothing #hide
+
+# ## Posterior
+#
+# To construct the posterior, we need to first build a `FiniteGP`, which
+# represents the infinite-dimensional GP at a finite number of input features:
+
+function build_finite_gp(θ)
+    f = build_gp_prior(θ)
+    return f(xtrain)
```
willtebbutt, line 120:
```suggestion
    return f(xtrain, θ.noise_scale^2)
```

examples/mauna-loa/script.jl
```diff
@@ -0,0 +1,236 @@
...
+#md nothing #hide
+
+# This allows us to write a function that, given the nested tuple of parameter values, constructs the GP prior:
+
+function build_gp_prior(θ)
+    k_smooth_trend = SE(θ.se1)
+    k_seasonality = Per(θ.per) * SE(θ.se2)
+    k_medium_term_irregularities = RQ(θ.rq)
+    k_noise_terms = SE(θ.se3) + θ.noise_scale^2 * WhiteKernel()
+    kernel = k_smooth_trend + k_seasonality + k_medium_term_irregularities + k_noise_terms
```
willtebbutt, line 108:
```suggestion
    k_noise_terms = SE(θ.se3)
    kernel = k_smooth_trend + k_seasonality + k_medium_term_irregularities
```

The reason for suggesting this (and moving the noise to the FiniteGP) is that the semantics are subtley different, and the thing I'm proposing is what I _think_ we want. The `WhiteKernel` says "non zero covariance between f(x) and f(y) if x == y", while noise-in-FiniteGP says "cov(f(x), f(y)) = 0 always". The latter means that we can always safely add a diagonal matrix to a kernel matrix, while the former requires you to carefully check that no two inputs used to construct the kernel matrix are the same, and to add off-diagonal terms if they are the same.

I would rather that we encourage users to use the FiniteGP way of handling noise, because it's typically what we actually want.

(Also, the gradient w.r.t. inputs isn't defined for the kernel matrix associated with the `WhiteKernel`, but it works just fine with noise-in-FiniteGP)

examples/mauna-loa/script.jl
```diff
@@ -0,0 +1,236 @@
...
+
+fpost_init = build_posterior_gp(ParameterHandling.value(θ_init))
+
+# Let's visualize what the GP fitted to the data looks like, for the initial choice of kernel hyperparameters.
+
+let
+    ## The `let` block creates a new scope, so any utility variables we define in here won't leak outside.
+    ## The return value of this block is given by its last expression.
+    plotdata()
+    plot!(fpost_init(1920:0.2:2030); ribbon_scale=2, label="posterior f(⋅)")  ## this returns the current plot object
```
willtebbutt, line 149: Could you drop the linewidth and alpha here? At least on my screen, it's hard to see the data underneath the posterior marginals.

willtebbutt: Or maybe provide two plots -- this one, and one which zooms in on a bit of the graph?

examples/mauna-loa/script.jl
```diff
@@ -0,0 +1,236 @@
...
+
+# This is the kernel with the point-estimated hyperparameters:
+
+fpost_opt.prior.kernel
+
+# And, finally, we can visualize our optimized posterior GP:
+
+let
+    plotdata()
+    plot!(fpost_opt(1920:0.2:2030); ribbon_scale=2, label="optimized posterior f(⋅)")
```
willtebbutt, line 235: Same here re visibility of the plot.

willtebbutt (commented):
This is great. I've left a couple of cosmetic remarks -- I think the thing I'm most interested in is discussing the use of a `WhiteKernel` vs noise-in-FiniteGP.

examples/mauna-loa/script.jl
st--: can you check the deployed docs preview ? I was trying to make it look decent there

examples/mauna-loa/script.jl
st--: clearly should've made a function for this :sweat_smile:

examples/mauna-loa/script.jl
st--: That's actually how I had it originally, but looking at Rasmussen&Williams again I just wanted to follow their example more closely... it somewhat changes the visualisation: using the WhiteKernel, its variance increases the uncertainty when plotting the posterior GP. Using the FiniteGP observation noise, we'd have to manually add the noise variance (`predict_y` in GPflow... was wondering if we want to support that in AbstractGPs directly, kinda goes more in the LatentGP direction though?)

Let me rephrase my understanding of the difference between WhiteKernel and noise-in-FiniteGP: in the former, covariance is finite only if xs[i] == xs[j] (based on the _values_ of the features), whereas in the latter, covariance is finite only for i == j (based on the _indices_). Does that match yours?

examples/mauna-loa/script.jl
st--: have a look now, better?
I think figuring out how to get two plots to look nice is overkill, but happy for you to improve that later:)

examples/mauna-loa/script.jl
willtebbutt: > Let me rephrase my understanding of the difference between WhiteKernel and noise-in-FiniteGP: in the former, covariance is finite only if xs[i] == xs[j] (based on the values of the features), whereas in the latter, covariance is finite only for i == j (based on the indices). Does that match yours?

Yes. This matches mine. I also think of it as always getting "fresh randomness" with the noise-in-FiniteGP version.

> Using the FiniteGP observation noise, we'd have to manually add the noise variance

Yes, we would have to do this. I think it's worth doing to emphasise that our abstractions make this particular distinction.

> (predict_y in GPflow... was wondering if we want to support that in AbstractGPs directly, kinda goes more in the LatentGP direction though?)

I wouldn't be overly keen on having a `predict_y` function -- I would rather just ensure that `LatentGP`s with simple Gaussian likelihoods work nicely and are an option that people can easily use.

st--: So what would you tell someone (like this notebook) how to easily get the observation uncertainty included?

Maybe in some cases using the WhiteKernel like this _is_ the right thing to do? I'd be happy to add a paragraph to the notebook explaining the distinction...

> (Also, the gradient w.r.t. inputs isn't defined for the kernel matrix associated with the WhiteKernel, but it works just fine with noise-in-FiniteGP)

I just realised I'm not sure what you mean and/or why this is...

willtebbutt: > So what would you tell someone (like this notebook) how to easily get the observation uncertainty included?

I would tell them to pick out the marginals they're interested in in `fpost_opt` by writing `fpost_opt(x, noise)` and to plot that. It's obviously not as clean as the current thing, but I think it makes sense.

> Maybe in some cases using the WhiteKernel like this is the right thing to do? I'd be happy to add a paragraph to the notebook explaining the distinction...

That would be great.

> I just realised I'm not sure what you mean and/or why this is...

I mean that `d/dx k(x, y)` doesn't really exist when `x == y` if `k` is a `WhiteKernel`.

examples/mauna-loa/script.jl
```diff
@@ -0,0 +1,243 @@
...
+# features:
+
+function build_finite_gp(θ)
+    f = build_gp_prior(θ)
+    return f(xtrain)
+end
+#md nothing #hide
+
+# In this notebook, we already included observation noise through the
+# `WhiteKernel` as part of the GP prior.
```
st--, line 126: Can we get cross-package doc references ?... I couldn't find anything with a quick google :(

examples/mauna-loa/script.jl
```diff
@@ -0,0 +1,243 @@
...
+#
+# The L-BFGS parameters were chosen because they seem to work well empirically.
+# You could also try with the defaults.
+
+default_optimizer = LBFGS(;
+    alphaguess=Optim.LineSearches.InitialStatic(; scaled=true),
+    linesearch=Optim.LineSearches.BackTracking(),
+)
+
+function optimize_loss(loss, θ_init; optimizer=default_optimizer, maxiter=1_000)
```
st--, line 185: @willtebbutt would you want to change this to `optimise_loss` [as well](https://github.com/JuliaGaussianProcesses/ApproximateGPs.jl/pull/71#discussion_r745326520), then? Seems somewhat confusing :S

willtebbutt: Nah, my position has changed on this now that I've slept on it. I think we probably ought to go with American English, so this is fine to my mind (see my most recent comment on that thread).

examples/mauna-loa/script.jl
```diff
@@ -0,0 +1,261 @@
...
+
+# This is the kernel with the point-estimated hyperparameters:
+
+fpost_opt.prior.kernel
+
+# Let's print the optimized values of the hyperparameters in a more helpful format:
+
+using Printf
+
+function show_params(nt::NamedTuple, pre=0)
```
st--, line 242: again, something that'd be nice to have in a utility file somewhere, so move it out of this notebook. where would be a good place for such things, though?

examples/mauna-loa/script.jl
```diff
@@ -0,0 +1,287 @@
...
+# which represents the infinite-dimensional GP at a finite number of input
+# features:
+
+function build_finite_gp(θ)
+    f = build_gp_prior(θ)
+    return f(xtrain)
+end
+#md nothing #hide
+
+# !!! info "`WhiteKernel` vs `FiniteGP` observation noise"
```
st--, line 125: @willtebbutt there you go. happy now? :smile:

examples/mauna-loa/script.jl
```diff
@@ -0,0 +1,287 @@
...
+#
+#     Moreover, the variance (uncertainty) of the posterior predictions
+#     includes the variance from the `WhiteKernel`, but does not include the
+#     variance of the observation noise passed to the `FiniteGP`. To include
+#     the observation noise in posterior predictions from the second approach,
+#     call `fpost_opt(xtest, noise_scale^2)`.
+#
+# !!! tip
+#     In practice, we recommend that you pass in observation noise to the
+#     `FiniteGP`, and omit the explicit `WhiteKernel`.
```
st--, line 151: @willtebbutt can you add a suggestion to clarify/motivate _why_ we would recommend this ?

willtebbutt:
```suggestion
#     In practice, we recommend that you pass in observation noise to the
#     `FiniteGP`, and omit the explicit `WhiteKernel`. This is because it is more efficient
#     (no need to check if `xtrain[i] == xtrain[j]`, if you wish to compute the gradient of
#     the log marginal probability w.r.t. `x` it won't work with the `WhiteKernel`.
```

Something like this?

willtebbutt (approved):
I'm happy with this once you are @st-- . Thanks for doing this!

examples/mauna-loa/script.jl
theogf: I think there's an issue with your sentence (and a missing parenthesis)

willtebbutt: Sorry, should be fixed now.

examples/mauna-loa/CO2_data.csv
devmotion: OK, I'll see if I can come up with a suggestion or fix it in a follow-up PR (which would, however, mean that the updates and workarounds to the build script will be part of this PR...). Where did you obtain the data from? It seems the data is different from the original data from https://gml.noaa.gov/ccgg/trends/data.html (Direct link: https://gml.noaa.gov/webdata/ccgg/trends/co2/co2_mm_mlo.csv) (You can e.g. just compare the time points of the first measurements and their values).

examples/mauna-loa/script.jl
```diff
@@ -0,0 +1,287 @@
...
+        if typeof(v) <: NamedTuple
+            res *= join(fill(" ", pre)) * "$(s):\n" * show_params(v, pre + 4)
+        else
+            res *= join(fill(" ", pre)) * "$s = $(@sprintf("%.3f", v))\n"
+        end
+    end
+    return res
+end
+
+print(show_params(ParameterHandling.value(θ_opt)))
```
devmotion, line 280: So you just want to print it? Then it would be better to just print it instead of creating a string in the first place.

examples/mauna-loa/script.jl
```diff
@@ -0,0 +1,287 @@
...
+
+# And, finally, we can visualize our optimized posterior GP:
+
+let
```
devmotion, line 284: Again no variables in this scope :shrug:

examples/mauna-loa/CO2_data.csv
st--: I copied the data from https://github.com/STOR-i/GaussianProcesses.jl/tree/master/notebooks. Yes, I noticed the discrepancy as well - not sure why that is. It says the scale changed, so that would explain a change in values, but not in the time points...

examples/mauna-loa/script.jl
st--: Can you come up with more convincing arguments? I find the claim "to be more efficient" very tenuous. Of all the contributions to the kernel matrix, the cost of evaluating the WhiteKernel is by far the smallest. I also don't see how computing the gradient w.r.t. x of the log marginal probability is a realistic use-case - maybe if you want to predict gradients of f _at training points_?

I'm happy to explain the pros & cons and to say "if in doubt, just use the FiniteGP way of adding observation noise", but I'm not yet convinced by actually making it a blanket _recommendation_...

examples/mauna-loa/script.jl
st--: print() doesn't work (out of the box) in Pluto. so ideally I would want something that implements show() appropriately...

examples/mauna-loa/script.jl
```diff
@@ -0,0 +1,287 @@
...
+    res = ""
+    for (s, v) in pairs(nt)
+        if typeof(v) <: NamedTuple
+            res *= join(fill(" ", pre)) * "$(s):\n" * show_params(v, pre + 4)
+        else
+            res *= join(fill(" ", pre)) * "$s = $(@sprintf("%.3f", v))\n"
+        end
+    end
+    return res
+end
```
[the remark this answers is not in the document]
st--: Could you provide a bit more background (happy to be linked to a decent resource) of why you would prefer this? The [Julia docs on string concatenation](https://docs.julialang.org/en/v1/manual/strings/#man-concatenation) don't mention this at all, they just say to use `*` for concatenation.

Why would this be more efficient? How much more efficient would it be - and does that really matter relative to the cost of actually optimising the model? Are there other benefits (or perhaps some disadvantages/limitations) of your proposed approach?

examples/mauna-loa/script.jl
```diff
@@ -0,0 +1,287 @@
...
+    fx = build_finite_gp(θ)
+    return posterior(fx, ytrain)
+end
+#md nothing #hide
+
+# Now we can put it all together to obtain a [`PosteriorGP`](@ref).
+# The call to `ParameterHandling.value` is required to replace the constraints
+# (such as `positive` in our case) with concrete numbers:
+
+fpost_init = build_posterior_gp(ParameterHandling.value(θ_init))
```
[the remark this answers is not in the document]
st--: Yes, better printing of kernels would be great. I think there should be two forms - basically a single-line one that's close to what you would write to create it in the first place, what `repr` should be, and then a nice multi-line one that works well in REPL or Pluto/Jupyter notebooks...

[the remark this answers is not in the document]
st--: Having looked through #244 again there doesn't actually seem to be much disagreement to me in it. We agreed on having both one-line and long-form output, and for the long-form output to contain some amount of nesting, probably indented by two spaces not a full tab. Whether you only nest the top-most level or recursively should be a very simple change, and maybe even something we can make configurable?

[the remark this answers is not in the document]
st--: I do still want some way of printing things in terms of signal standard deviations (not variances) and lengthscales (not inverse lengthscale scales), which is generally more immediately interpretable...

examples/mauna-loa/CO2_data.csv
devmotion: Yes, exactly, the time points are the really weird discrepancy.

examples/mauna-loa/script.jl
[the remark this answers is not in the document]
devmotion: Because it is expensive to create all these intermediate strings. Or arrays of string with `fill`, which are then `join`ed, just to be printed or written to some other I/O device.

It is mentioned in the Julia documentation in the performance section that one should avoid such intermediate strings: https://docs.julialang.org/en/v1/manual/performance-tips/index.html#Avoid-string-interpolation-for-I/O

We also fixed this in e.g. KernelFunctions (the initial version also concatenated strings a lot).

examples/mauna-loa/script.jl
devmotion: But that's a Pluto weirdness :shrug: For all kind of standard terminal output and functionality one has to use `with_terminal() do ... end` in Pluto, just because it hacks into the display machinery and has a very opinionated way of displaying stuff.

examples/mauna-loa/script.jl
```diff
@@ -0,0 +1,302 @@
...
+#     so any utility variables we define in here won't leak outside. This is
+#     particularly helpful to keep notebooks tidy! The return value of this
+#     block is given by its last expression.
+
+let
+    plotdata()
+    label = "posterior f(⋅)"
+    plot_gp!(fpost_init; label)  # this returns the current plot object
+end  # and so the plot object will be shown
+## `label` is not available outside the `let` scope
```
devmotion, line 191: :smile: You just added this artificial variable so the comment is somewhat more reasonable, didn't you? I really don't think this makes sense at all. `let` can be useful for the reasons described in the comment but there's no reason for a block here - I think it's even worse with this artificial variable since I don't think anyone would write code in this way.

It is also a bit unfortunate to use a syntax here (`; label)`) that only works with more recent versions of Julia (IIRC Julia >= 1.4).

Please just replace this with the following lines:
```suggestion
plotdata()
plot_gp!(fpost_init; label="posterior f(⋅)")
```

examples/mauna-loa/script.jl
```diff
@@ -0,0 +1,302 @@
...
+
+print(show_params(ParameterHandling.value(θ_opt)))
+
+# And, finally, we can visualize our optimized posterior GP:
+
+let
+    plotdata()
+    label = "optimized posterior f(⋅)"
+    plot_gp!(fpost_opt; label)
+end
```
devmotion, line 302: Same here, this is so unidiomatic that I really don't think it should be used in a tutorial (additionally, as mentioned above `; label)` is only supported in new versions). Please just use
```suggestion
plotdata()
plot_gp!(fpost_opt; label="optimized posterior f(⋅)")
```

devmotion (dismissed):

examples/mauna-loa/script.jl
willtebbutt: > Of all the contributions to the kernel matrix, the cost of evaluating the WhiteKernel is by far the smallest.

Here's an example where it's non-trivial:

```julia
using KernelFunctions, BenchmarkTools, LinearAlgebra
x = randn(1_000);

julia> @benchmark kernelmatrix(SEKernel(), $x) + kernelmatrix(WhiteKernel(), $x)
BenchmarkTools.Trial: 370 samples with 1 evaluation.
 Range (min … max):  10.022 ms … 20.796 ms  ┊ GC (min … max):  0.00% … 20.08%
 Time  (median):     14.022 ms              ┊ GC (median):     6.03%
 Time  (mean ± σ):   13.537 ms ±  2.690 ms  ┊ GC (mean ± σ):  17.65% ± 15.29%

      █▄
  ▂▂▂▆██▅▄▃▃▃▁▂▃▂▃▂▂▁▁▂▁▁▂▃▁▃▄▄▄▄▄▄▅▄▃▄▄▄▃▄▂▃▃▃▃▃▂▃▃▂▃▂▁▂▃▁▁▂ ▃
  10 ms           Histogram: frequency by time        19.3 ms <

 Memory estimate: 24.82 MiB, allocs estimate: 17.

julia> @benchmark kernelmatrix(SEKernel(), $x) + Diagonal(fill(1.0, length($x)))
BenchmarkTools.Trial: 422 samples with 1 evaluation.
 Range (min … max):   8.995 ms … 17.243 ms  ┊ GC (min … max):  0.00% … 33.75%
 Time  (median):     12.878 ms              ┊ GC (median):     6.05%
 Time  (mean ± σ):   11.834 ms ±  2.342 ms  ┊ GC (mean ± σ):  18.83% ± 16.00%

    ▂█▆▃                          ▂▁▂▃▁▂▄▃▂ ▁ ▁  ▁▁
  ▄▅████▇▇▅▁▄▁▅▄▁▁▁▁▁▁▁▄▁▁▁▁▁▄▄▁▅▇█████████████▇▇██▇▅█▆▅▅▄▄▄▅ ▇
  8.99 ms      Histogram: log(frequency) by time      16.3 ms <

 Memory estimate: 22.91 MiB, allocs estimate: 11.
```

> I also don't see how computing the gradient w.r.t. x of the log marginal probability is a realistic use-case - maybe if you want to predict gradients of f at training points?

There are a couple of good examples here. The first is gradient-based optimisation of pseudo-input locations:
```julia
using AbstractGPs, KernelFunctions, BenchmarkTools, Zygote
x = randn(1_000);
y = randn(1_000);
function elbo(z)
    f = GP(SEKernel() + WhiteKernel())
    return AbstractGPs.elbo(VFE(f(z)), f(x), y)
end
z = range(-4.0, 4.0; length=11);
Zygote.gradient(elbo, z)
```
yields
```julia
([2.877651767042302e17, 1.1193746344481599e18, 5.584065115922174e18, 1.4266911109886681e19, 1.293142891376751e19, 3.147369002799461e17, -1.4651782340916806e19, -1.2866523606484107e19, -5.623767597697787e18, -1.6447138973739287e18, -1.5529275658874477e17],)
```
The other is if you want to apply an input transformation before hitting the GP, and optimise it (the following is pseudo-code, doesn't actually work):
```julia
function objective(x, theta)
    new_x = parametrised_function(theta, x)
    f = GP(SEKernel() + WhiteKernel())
    return logpdf(f(new_x), y)
end
```

examples/mauna-loa/script.jl
```diff
@@ -0,0 +1,287 @@
...
+
+# This allows us to write a function that, given the nested tuple of parameter values, constructs the GP prior:
+
+function build_gp_prior(θ)
+    k_smooth_trend = SE(θ.se_long)
+    k_seasonality = Per(θ.seasonal.per) * SE(θ.seasonal.se)
+    k_medium_term_irregularities = RQ(θ.rq)
+    k_noise_terms = SE(θ.se_short) + θ.noise_scale^2 * WhiteKernel()
+    kernel = k_smooth_trend + k_seasonality + k_medium_term_irregularities + k_noise_terms
+    return GP(kernel)  # [`ZeroMean`](@ref) mean function by default
```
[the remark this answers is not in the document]
st--: Hm, `documenter=true` ends up leaving all the `nothing #hide` lines in the rendered docs. And the ``[`FiniteGP`](@ref)`` still doesn't work with it anyways... :(

examples/mauna-loa/script.jl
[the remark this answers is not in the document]
st--: The version with the IOBuffer is about 2x faster & 2x less allocations (though I wouldn't call it "expensive" either way - it's a few kilobytes and around 100 microseconds...).

I'm glad you showed me a better way of doing it - I would not have made the connection based on the official docs you linked to (that very much sounds like it's just for file i/o, not for constructing strings in general)...

examples/mauna-loa/script.jl
```diff
@@ -0,0 +1,301 @@
...
+
+show_params(nt::Union{Dict,NamedTuple}) = String(take!(show_params(IOBuffer(), nt)))
+function show_params(io, nt::Union{Dict,NamedTuple}, indent::Int=0)
+    for (s, v) in pairs(nt)
+        if v isa Union{Dict,NamedTuple}
+            println(io, " "^indent, s, ":")
+            show_params(io, v, indent + 4)
+        else
+            println(io, " "^indent, s, " = ", @sprintf("%.3f", v))
+        end
```
devmotion, line 291: There are some avoidable string concatenations left (`" "^indent` and `@sprintf` will allocate strings with whitespace and not print them directly to `io`) that one can get rid of. To reduce the amount of code I moved the identical operations out of the `if` block:
```suggestion
        for _ in 1:indent
            print(io, " ")
        end
        print(io, s)
        if v isa Union{Dict,NamedTuple}
            println(io, ":")
            show_params(io, v, indent + 4)
        else
            println(io, " = ")
            @printf(io, "%.3f", v)
        end
```

st--: Yes, this would be marginally faster, but I found the version as it is in the script now rather more readable, and the impact on performance is negligible.

st--: It's good to write performant code when that's easy, but speed doesn't trump everything else, particularly not when it's a few percent shaved off a function that's nowhere near the bottleneck and already running in microseconds not milliseconds..

devmotion: Sure, this is not completely satisfying for a tutorial (even though I think the more verbose version is more readable :shrug:) - but IMO neither of the two versions is. But I assumed this will be moved somewhere else soon, and hence I thought it would be better to not having to think about it and optimize it again when it's moved.

st--: Yes, this should get moved elsewhere so it's easy to re-use wherever! I'd still rather keep it like this though.

examples/mauna-loa/script.jl
willtebbutt: I'm happy with the updated advice, so resolving.

willtebbutt (approved):
Re-approving following some changes -- very happy to see this merged

PR #243, opened 2021-11-09
(Do not merge) **BREAKING** Remove LatentGP

willtebbutt:
**Summary**

All of our work involving the LatentGP and friends type now resides in ApproximateGPs.jl. Moreover, literally nothing in this package uses that type. Given that we're going to need to iterate on the LatentGP type a bit (following last week's discussions about its shortcomings), I propose to move it to ApproximateGPs (see counterpart PR there).

**Proposed changes**

Move LatentGP and friends to ApproximateGPs.jl

* ...

I considered not moving it, and trying to evolve the LatentGP abstractions while adding additional code to ApproximateGPs, but my feeling is that's going to hinder development, and force us to make lots of breaking releases of AbstractGPs (which are a bit of a pain for downstream packages to process, and are annoying if none of the change affect those packages).

**Breaking changes**

Removal of LatentGP. I doubt this will affect anyone but us, as I don't think anyone is using it.


st--:
Hm, on the one hand I agree it's nice if it's in the place as where it gets used - maybe the following is too hypothetical: I'm just wondering if in the longer run that'll make it harder to get it as a base for multiple packages - maybe someone else wants to implement some other approximations, wouldn't it then be nice if they could depend on just AbstractGPs? Or if someone wanted to do MCMC on a LatentGP model (I _think_ that should be doable with just `LatentGP`? makes me think we ought to have a notebook for that :smile: )

willtebbutt:
Yeah, that's a good point, and I guess was the original motivation for having it in this package.

Lets leave this open -- I'm currently implementing a LatentGP that requires entire sample paths as part of some of my research, so I'm going to be proposing some changes to LatentGP. Lets see how that process goes, and revisit if it's truly awful. 🤞 it'll be fine.

simsurace:
Is this still something worth pursuing?

PR #253, opened 2021-11-16
Quickfix: add missing compat entries

st--:
(and a missed edit from #240)

Supersedes #245, #246, #247, #248, #249, #250, #251, #252.


willtebbutt (approved):
It's always amusing when compathelper does this. Thanks for fixing.

st--:
Well, I opened an issue here: https://github.com/JuliaRegistries/CompatHelper.jl/issues/403 ...

Issue #242, opened 2021-11-09
Incomplete implementation of <: AbstractMvNormal


theogf:
If we actually define `FiniteGP <: AbstractMvNormal` we should follow the API defined in https://github.com/JuliaStats/Distributions.jl/blob/master/src/multivariate/mvnormal.jl, i.e.

- [x] `length(d)` :        vector dimension
- [x] `mean(d)`:          the mean vector (in full form)
- [x] `cov(d)`:           the covariance matrix (in full form)
- [ ] `invcov(d)`:        inverse of covariance
- [ ] `logdetcov(d)`:     log-determinant of covariance
- [ ] `sqmahal(d, x)` :        Squared Mahalanobis distance to center
- [ ] `sqmahal!(r, d, x)`:    Squared Mahalanobis distances
- [ ] `gradlogpdf(d, x)`:     Gradient of logpdf w.r.t. x
- [x] `_rand!(d, x)`:         Sample random vector(s)
- [ ] `params(d)`

willtebbutt:
Yeah, we should probably implement these things. Does Distributions provide any API testing suites that we can utilise?

devmotion:
Of course, not :smile: I always wanted to have some though (for all kinds of distributions), so a PR would be great :stuck_out_tongue:

willtebbutt:
Also, are there fallback definitions for any of these that we get for free?

theogf:
Haha I was going to say the same. It seems it's already there in the tests though:
https://github.com/JuliaStats/Distributions.jl/blob/master/test/mvnormal.jl

theogf:
> Also, are there fallback definitions for any of these that we get for free?

Only `sqmahal` can fallback on `sqmahal!`

willtebbutt:
> Haha I was going to say the same. It seems it's already there in the tests though:

Nice. @devmotion would you be open to a PR that pretty much just moves that function into `src`? (Into a submodel called `TestUtils` or whatever)

devmotion:
In principle, it would be possible to define fallbacks for `logdetcov` and `invcov` as well :thinking: I don't know why they are not included.

theogf:
> > Haha I was going to say the same. It seems it's already there in the tests though:
> 
> Nice. @devmotion would you be open to a PR that pretty much just moves that function into `src`? (Into a submodule called `TestUtils` or whatever)

[...]

PR #254, opened 2021-11-23
Implement AbstractMvNormal API for FiniteGP

theogf:
**Summary**
Since we defined `FiniteGP` as a subtype of `AbstractMvNormal` in #123, we need to implement the required API as described in #242.

**Proposed changes**
This PR implements all the needed functions and tests the API functionality using the `TestUtils` module from `Distributions.jl`.
I was also able to merge the `logpdf` function for `AbstractVector` and `AbstractMatrix` without needing a reshape.

For a lot of functions I added an optional argument to give the mean/covariance directly if it was already computed as it is often the case in the use of these functions

I also added `tr_Xt_invA_X` which was missing from our common operations.

We also might want to implement the problematic `eltype` which is often used in `Distributions.jl`, but then... what should it return...

:man_shrugging: 
**Breaking changes**
None hopefully


src/util/common_covmat_ops.jl
```diff
@@ -1,8 +1,9 @@
-# If a matrix is `Diagonal`, we generally don't need to wrap it in a `Symmetric`, because
+# If a matrix is `Diagonal` or if it is a `Cholesky` object, we generally don't need to wrap it in a `Symmetric`, because
```
willtebbutt, line 1: I think this is over the character limit. Not sure why the formatting tool didn't pick up on it.

src/finite_gp_projection.jl
```diff
@@ -300,15 +302,56 @@ true
...
     m, C_mat = mean_and_cov(f)
     C = cholesky(_symmetric(C_mat))
     T = promote_type(eltype(m), eltype(C), eltype(Y))
-    return -((size(Y, 1) * T(log(2π)) + logdet(C)) .+ diag_Xt_invA_X(C, Y .- m)) ./ 2
+    return -((size(Y, 1) * T(log2π) + logdetcov(f, C)) .+ sqmahal(f, Y, (m, C))) ./ 2
+end
+
+function Distributions.logdetcov(f::FiniteGP, C=cov(f))
+    return logdet(C)
+end
```
willtebbutt, line 316:
```suggestion
Distributions.logdetcov(f::FiniteGP, C=cov(f)) = logdet(C)
```

src/finite_gp_projection.jl
```diff
@@ -300,15 +302,56 @@ true
...
-Distributions.loglikelihood(f::FiniteGP, Y::AbstractMatrix{<:Real}) = sum(logpdf(f, Y))
+Distributions.loglikelihood(f::FiniteGP, Y::AbstractVecOrMat{<:Real}) = sum(logpdf(f, Y))
 
-function Distributions.logpdf(f::FiniteGP, Y::AbstractMatrix{<:Real})
+function Distributions.logpdf(f::FiniteGP, Y::AbstractVecOrMat{<:Real})
     m, C_mat = mean_and_cov(f)
     C = cholesky(_symmetric(C_mat))
     T = promote_type(eltype(m), eltype(C), eltype(Y))
-    return -((size(Y, 1) * T(log(2π)) + logdet(C)) .+ diag_Xt_invA_X(C, Y .- m)) ./ 2
+    return -((size(Y, 1) * T(log2π) + logdetcov(f, C)) .+ sqmahal(f, Y, (m, C))) ./ 2
```
willtebbutt, line 311: Not that it seems especially important, but does it make a difference using `logdetcov` here vs just `logdet`?

src/finite_gp_projection.jl
```diff
@@ -300,15 +302,56 @@ true
 """
 logpdf(f::FiniteGP, y::AbstractVecOrMat{<:Real})
 
-function Distributions._logpdf(f::FiniteGP, y::AbstractVector{<:Real})
-    return first(logpdf(f, reshape(y, :, 1)))
-end
-
-Distributions.loglikelihood(f::FiniteGP, Y::AbstractMatrix{<:Real}) = sum(logpdf(f, Y))
+Distributions.loglikelihood(f::FiniteGP, Y::AbstractVecOrMat{<:Real}) = sum(logpdf(f, Y))
```
devmotion, line 305: Maybe only define it for `AbstractMatrix`? For vectors it falls back to `logpdf` automatically. There's also a default non-allocating version for the matrix case but maybe the allocating one is faster here?

src/finite_gp_projection.jl
willtebbutt: Also, the `logdetcov` and `sqmahal` implementations just feel a bit weird to me. I get why you've allowed for the additional argument (in case you already have `m` and `C` lying around), but it's really not clear to me why we wouldn't just
1. stick with `logdet`, and
2. extend `diag_Xt_invA_X` to handle vectors gracefully.

Could you elaborate a bit on why you chose the design proposed in this PR?

src/finite_gp_projection.jl
```diff
@@ -300,15 +302,56 @@ true
...
 
-function Distributions.logpdf(f::FiniteGP, Y::AbstractMatrix{<:Real})
+function Distributions.logpdf(f::FiniteGP, Y::AbstractVecOrMat{<:Real})
```
devmotion, line 307: Is there a reason to implement `logpdf` instead of `_logpdf`, as expected by the interface?

willtebbutt (commented):
Looking forward to having this be resolved.

src/finite_gp_projection.jl
theogf: Done, for the matrix the allocating one is definitely faster, or one would need to rewrite `logpdf` all over again taking care of allocations.

src/finite_gp_projection.jl
theogf: Nope, changed it now

src/finite_gp_projection.jl
```diff
@@ -300,15 +302,56 @@ true
...
 logpdf(f::FiniteGP, y::AbstractVecOrMat{<:Real})
 
-function Distributions._logpdf(f::FiniteGP, y::AbstractVector{<:Real})
-    return first(logpdf(f, reshape(y, :, 1)))
-end
-
 Distributions.loglikelihood(f::FiniteGP, Y::AbstractMatrix{<:Real}) = sum(logpdf(f, Y))
 
-function Distributions.logpdf(f::FiniteGP, Y::AbstractMatrix{<:Real})
+function Distributions._logpdf(f::FiniteGP, Y::AbstractVecOrMat{<:Real})
```
willtebbutt, line 307: IIRC the reason for doing `logpdf` rather than `_logpdf` is because `_logpdf` isn't actually what is expected by Distributions for `Y::AbstractMatrix` -- it expects `_logpdf!` so that some pre-allocation can be done. Unless things have changed since I last looked...
@devmotion could you comment?

src/finite_gp_projection.jl
theogf: I somehow wanted to reuse the std Distributions functions but I agree that for `logdetcov` it feels a bit weird.
For `sqmahal` however it makes sense since it allows to have a scalar when `Y` is a vector and a vector when `Y` is a Matrix, which is what you also want from `logpdf`

willtebbutt: Hmm yeah. I agree that `sqmahal` probably has the correct semantics, it still feels odd to me that `sqmahal` has three arguments. I wonder whether it would make more sense to do something like
```julia
sqmahal((m, C), Y) = ...
sqmahal(fx::FiniteGP, Y) = sqmahal(mean_and_cov(fx), Y)
```
so that there's not redundant information floating around? ie. you really don't need `fx` and `(m, C)` at the same time.

theogf: Yes you're right, I defined `_sqmahal` for that

src/finite_gp_projection.jl
devmotion: No, the default implementation of every out-of-place method should only call `_logpdf(::MultivariateDistribution, ::AbstractVector)`. This was not always the case (maybe there are still some bugs) but it is the case in recent versions: https://github.com/JuliaStats/Distributions.jl/blob/9f494537197b41a383996a20d32bfe336d07068f/src/multivariates.jl#L236-L240

However, I assume that it could still be useful here to not use the default implementation of `logpdf(::FiniteGP, ::AbstractMatrix)` since it will iterate through the columns and hence probably be slower. Otherwise if you only implement `_logpdf` here it will never be called with matrices, regardless of whether the second argument allows `AbstractMatrix` or not.

willtebbutt: Ahh okay. Is there any disadvantage to implementing `logpdf` directly rather than `_logpdf` in general?

devmotion: Currently, in theory yes but in practice probably no: Internally, some functions call `_logpdf` directly (but not `logpdf`) if the size is checked already. This will change in https://github.com/JuliaStats/Distributions.jl/pull/1391/files#diff-c7d9a62e5d5a29aca07b2c1f1365565e127c595ec68316f1c1ac0da331b5c6b5 though: To avoid such problems, we will always call `logpdf` but add `@inbounds` if the size is guaranteed to be correct.

willtebbutt: Ace. I vote for `logpdf` then.

src/finite_gp_projection.jl
```diff
@@ -300,15 +302,36 @@ true
...
+function Distributions.gradlogpdf(f::FiniteGP, x::AbstractArray)
+    m, C = mean_and_cov(f)
+    return _gradlogpdf(m, C, x)
+end
+
+_gradlogpdf(m, C, x) = _symmetric(C) \ (m .- x)
+
+function Distributions.params(f::FiniteGP)
+    return (f.f, f.x, f.Σy)
 end
```
willtebbutt, line 337:
```suggestion
Distributions.params(f::FiniteGP) = (f.f, f.x, f.Σy)
```

src/finite_gp_projection.jl
theogf: Ok I switch back to logpdf then!

willtebbutt (approved):
Subject to CI passing, a patch bump, and the promised reversion to using `logpdf` rather than `_logpdf`, I'm happy for this to go in.

PR #260, opened 2022-01-19
**BREAKING**: makes GP evaluation explicitly require RowVecs/ColVecs wrapper; calling GP with AbstractMatrix now errors

### LINKED ISSUE #257: `GP`'s call should not accept matrix (without RowVecs/ColVecs)
It'd be better to raise an error (and explicitly point people to RowVecs/ColVecs) - it just took me fifteen minutes to figure out why the assert length(fx) == length(y) failed for me :facepalm:

st--:
Resolves #257


src/finite_gp_projection.jl
```diff
@@ -23,6 +23,17 @@ end
...
+    nrows, ncols = size(X)
+
+    if obsdim == 1
+        ndata, dim, xvecs = nrows, ncols, "RowVecs"
+    elseif obsdim == 2
+        ndata, dim, xvecs = ncols, nrows, "ColVecs"
+    else
+        throw(ArgumentError("obsdim must be 1 or 2"))
+    end
+    @info "Evaluating the GP on a $nrows × $ncols matrix, interpreted as $ndata points of dimension $dim. To remove this message, wrap the matrix in $xvecs."
```
st--, line 35:
```suggestion
    @info "Evaluating the GP on a $nrows × $ncols matrix, interpreted as $ndata points of dimension $dim. (To get rid of this message, wrap the matrix in $xvecs)."
```

theogf: Maybe add a reference to https://juliagaussianprocesses.github.io/KernelFunctions.jl/stable/api/#Input-Types

devmotion, line 35: I think we should either use `@debug`, show the message only in the case of an error, or properly deprecate support for matrices (would be inconsistent with KernelFunctions though). Otherwise we mainly annoy users with logs.

st--: I added such a reference to api.md (055f353). Do you also want it in this message?

st--: Showing only in case of error lets it work (but wrongly) when X is square (though that's probably a rare case). If any user doesn't want to see the message anymore, they can simply get rid of it by wrapping in ColVecs as suggested, no? We could be firmer on "we recommend using ColVecs(X) instead"...

devmotion: My point is: either we support it or we don't. But if we support it we should beither spam the users no cause slowdowns (logging is quite slow).

st--:
Somehow the docs preview doesn't seem to reflect the changes.

The PPL test failure is unrelated (Turing/libtask_jll version incompatiblities).

src/finite_gp_projection.jl
st--: I think it's nice to have it give a meaningful message, but still continue running. I believe that will make the transition easier for users coming from other packages. (Instead of outright deprecating passing a matrix, and eventually dropping support entirely.)

How about
```suggestion
    @info "For vector-of-vectors input represented by a matrix X, please use f(ColVecs(X)) or f(RowVecs(X)) instead (depending on your data layout). f(X) is now evaluating the GP on a $nrows × $ncols matrix, interpreted as $ndata points of dimension $dim."
```
?

devmotion: No, as a user I don't want to see messages if it's supported. There's not something like supporting it a bit, either it's part of the API or not.

If you want users to move to ColVecs etc then deprecating it is the right approach. We can still support it for some longer time, it doesn't mean we have to remove it in the next breaking release.

I'm fine with deprecating it if we agree on this and are consistent throughout different packages (thus eg requires changes in KernelFunctions as well).

st--: If you don't want to see messages, just use `f(ColVecs(X))`. What's the problem with offering people a way of doing it how they might naively expect it to work by just passing in a matrix, and getting an explanation of what's happening and what they should do instead?

devmotion: `@info` is not designed for this use case. This is exactly what deprecations are for.

st--: I disagree with that. The point here is to take something that a new user might expect to do (call `f(X)` with a matrix as they might expect from python or matlab GP codes), and tell them what they should be doing instead. `@deprecate` doesn't allow for any custom messages.

I'd like to hear what everyone else is thinking on this. @willtebbutt , @theogf ?

st--: Another issue with deprecations is that they won't print a message by default, and I wouldn't expect a new user to think of running `julia --depwarn=yes`. The point of this _is_ that by default you do get a message. I see it as a "make it easy for users to learn".

If we want to change it to always require obsdim, we'll have to change the APIs - in the current API you can't pass `obsdim` - `FiniteGP` itself isn't exported, the public API expects you to just call `f(X)`.

devmotion: This discussion doesn't seem to lead anywhere unfortunately :shrug: I've said everything I want to say already above.

> If we want to change it to always require obsdim, we'll have to change the APIs - in the current API you can't pass `obsdim` - `FiniteGP` itself isn't exported, the public API expects you to just call `f(X)`.

No, we wouldn't have to change the API. We would just add a dispatch on `X` in https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/blob/e7c67bd24c454f0b8179032ecbb82284887159c0/src/finite_gp_projection.jl#L38 and wrap matrices in `ColVecs` or `RowVecs` already at this stage (and probably deprecate the non-exported `FiniteGP` constructor with `X::AbstractMatrix`).

st--: We discussed this at the meeting today, and it was decided to make it error explicitly, so I'll resolve this thread now.

docs/src/api.md
```diff
@@ -48,6 +48,16 @@ fx = f(x, Σ)
...
+!!! note
+
+    When considering higher-dimensional input features, these are often
+    provided as a matrix `X`. Evaluating `f(X)` interprets these as a vector of
+    vectors, where by default each _column_ of `X` is seen to represent an
+    input point. The implementation is handled by KernelFunctions.jl; see its
+    [documentation on Input
+    Types](https://juliagaussianprocesses.github.io/KernelFunctions.jl/stable/api/#Vector-Valued-Inputs)
+    for more details.
+
```
willtebbutt, line 60: Presumably this needs to be removed or modified?

src/finite_gp_projection.jl
```diff
@@ -20,10 +20,15 @@ function FiniteGP(f::AbstractGP, x::AbstractVector, σ²::Real=default_σ²)
...
-)
-    return FiniteGP(f, KernelFunctions.vec_of_vecs(X; obsdim=obsdim), σ²)
+function FiniteGP(f::AbstractGP, X::AbstractMatrix, σ²=default_σ²)
+    nrows, ncols = size(X)
+    throw(ArgumentError("""projecting a GP on a finite number of points only accepts vectors (e.g. a vector of vectors for multi-dimensional features), but was called with a $nrows × $ncols matrix.
+- If this is supposed to represent $ncols data points of dimension $nrows,
+  wrap it in ColVecs (e.g. `f(ColVecs(X), σ²)`).
+- If this is supposed to represent $nrows data points of dimension $ncols,
+  wrap it in RowVecs (e.g. `f(RowVecs(X), σ²)`).
+For more details, see https://juliagaussianprocesses.github.io/KernelFunctions.jl/stable/api/#Input-Types
```
willtebbutt, line 32: Would it make sense to add a performance-related note here? Something to reassure users that `ColVecs` and `RowVecs` are just wrappers around the matrix, so the overall performance isn't affected.

willtebbutt (commented):
A couple of thoughts

willtebbutt:
Oo, also -- could you add a small `@test_throws` test somewhere, to help ensure that we don't accidentally remove the informative error message at some point in the future?

docs/src/concrete_features.md
```diff
@@ -1,26 +1,6 @@
 # Features
 
-## Plotting
```
devmotion, line 3: These changes seem unrelated, would have been cleaner to make them in a separate PR.

devmotion (commented):
I think the PR is not completely ready yet but it's definitely better to decide and state clearly that matrices are not supported instead of allowing them but still printing a log message.

I have the following questions:
- Was there agreement that matrices should not be allowed in KernelFunctions either (I would be fine with it but I know that @theogf wanted to support matrices in the past)? I think this PR should only be merged after the behaviour in KernelFunctions is changed, and it should not be merged at all if we want to continue supporting matrices in KernelFunctions. It seems you also link to the documentation of KernelFunctions in the docstring but currently this documentation explains why matrices are supported: https://juliagaussianprocesses.github.io/KernelFunctions.jl/stable/design/#Why-We-Have-Support-for-Both
- Have you considered deprecating the constructor (possibly with a custom deprecation message) and removing it in the next breaking release instead of throwing an error? Throwing an error demands a breaking release but in this case we could just remove the constructor directly.
- Have you considered handling the matrices in `(f::AbstractGP)(x...)` instead of (or in addition) to the `FiniteGP` constructor? `f(x...)` seems to be the official API and hence it is a bit surprising to only error in an internal method - but to refer to `f(x...)` in the error message.

theogf:
> Was there agreement that matrices should not be allowed in KernelFunctions either (I would be fine with it but I know that @theogf wanted to support matrices in the past)? I think this PR should only be merged after the behaviour in KernelFunctions is changed, and it should not be merged at all if we want to continue supporting matrices in KernelFunctions. 

I don't care anymore if we support matrices or not in KernelFunctions. I think this is a useful feature, but if we want to remove it it's fine. However, I disagree with the order of deprecation. We should deprecate AbstractGPs first as it has zero impact on KernelFunctions.jl. KF is still a very distinct part of the whole ecosystem anyway

devmotion:
> We should deprecate AbstractGPs first as it has zero impact on KernelFunctions.jl. KF is still a very distinct part of the whole ecosystem anyway

It's distinct but in both cases currently one can use `ColVecs` etc and matrices, so dropping support for matrices in only one of the packages is inconsistent. Both are part of the GP ecosystem and KernelFunctions defines and explains `ColVecs` etc. IMO the inconsistency becomes really apparent by the fact that the KernelFunctions docs - which are linked from the docstring in this PR - explicitly explain why matrices are supported as well.

As a middle ground, we could just remove the default value for `obsdim` (I mentioned this in a thread above). Eg Distances did the same, and this would avoid any confusion and implicit assumptions about whether rows or columns are observations.

devmotion:
I think this PR should not be merged in light of the (soon to be merged) `obsdim` changes. Instead I think it would be more consistent to require an explicit `obsdim` argument as well for matrices.

devmotion:
https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/pull/264 was merged, so I think this PR can be closed.

PR #264, opened 2022-01-31
Deprecate implicit `obsdim` keyword argument and improve tests

### LINKED ISSUE #257: `GP`'s call should not accept matrix (without RowVecs/ColVecs)

devmotion:
Pendant to https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/429.

The PR
- widens the type annotation of the `obsdim` keyword argument such that the new default `nothing` in KernelFunctions works
- moves the `AbstractMatrix` handling to the user-facing API such that internally one only has to deal with vectors such as `ColVecs` and `RowVecs`
- improves tests by testing the user-facing API instead of internal constructors
- adds tests of deprecated implicit `obsdim` keyword argument

Fixes #257. Alternative to #260.


test/finite_gp_projection.jl
```diff
@@ -235,8 +246,10 @@ end
 end
 
 @testset "Docs" begin
-    docstring = string(Docs.doc(logpdf, Tuple{FiniteGP,Vector{Float64}}))
-    @test occursin("logpdf(f::FiniteGP, y::AbstractVecOrMat{<:Real})", docstring)
+    docstring = string(Docs.doc(logpdf, Tuple{AbstractGPs.FiniteGP,Vector{Float64}}))
+    @test occursin(
+        "logpdf(f::AbstractGPs.FiniteGP, y::AbstractVecOrMat{<:Real})", docstring
```
devmotion, line 251:
```suggestion
        "logpdf(f::FiniteGP, y::AbstractVecOrMat{<:Real})", docstring
```

src/finite_gp_projection.jl
```diff
@@ -36,6 +30,11 @@ end
 Base.length(f::FiniteGP) = length(f.x)
 
 (f::AbstractGP)(x...) = FiniteGP(f, x...)
+function (f::AbstractGP)(
+    X::AbstractMatrix, args...; obsdim::Union{Int,Nothing}=KernelFunctions.defaultobs
```
theogf, line 34: why the `args...` instead of `sigma^2` here?

src/finite_gp_projection.jl
```diff
@@ -36,6 +30,11 @@ end
 Base.length(f::FiniteGP) = length(f.x)
 
 (f::AbstractGP)(x...) = FiniteGP(f, x...)
```
theogf, line 32: Should it be 
```suggestion
(f::AbstractGP)(x...; kwargs...) = FiniteGP(f, x...; kwargs...)
```
or is it done automatically?

devmotion: No, it isn't done automatically (and hence one can't change `obsdim` with the official API - but this is not tested either...). However, since `FiniteGP` does not accept keyword arguments it is not really needed currently after applying the fixes and dispatches in this PR.

src/finite_gp_projection.jl
devmotion: To capture all possible arguments (if someone decides to create some other internal `FiniteGP` constructors) and to not have to worry about default values of some arguments. Similar to what we do in KernelFunctions, with this PR we just forward everything to the vector case and handle arguments, default values, etc. there.

src/finite_gp_projection.jl
```diff
@@ -36,6 +30,11 @@ end
...
+function (f::AbstractGP)(
+    X::AbstractMatrix, args...; obsdim::Union{Int,Nothing}=KernelFunctions.defaultobs
+)
+    return FiniteGP(f, KernelFunctions.vec_of_vecs(X; obsdim=obsdim), args...)
```
devmotion, line 36: Even better in case someone decides to implement `(f::T)(args...)` for some `T<:AbstractGP` in a different way:
```suggestion
    return f(KernelFunctions.vec_of_vecs(X; obsdim=obsdim), args...)
```

theogf (approved):

theogf:
Just need another patch bump

devmotion:
Done :slightly_smiling_face:

Issue #258, opened 2022-01-11
elbo <= logpdf in test utils


willtebbutt:
This intersting thing occured in LinearMixingModels.jl:
https://github.com/invenia/LinearMixingModels.jl/issues/44

My suspicion is that we should just add a bit of jitter to `f(z)` in this line: https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/blob/e7c67bd24c454f0b8179032ecbb82284887159c0/src/util/test_util.jl#L207

The net result will be to make the approximation a little bit worse, hopefully generally avoiding these problems.

I'll try to concoct a MWE that reproduces the error that we can include in this package.

PR #265, opened 2022-02-01
TestUtil Jitter

willtebbutt:
**Summary**

In our test utils, we currently compute covariance matrices that need to be inverted under the default jitter, 1e-18.
This occassionally makes things not invertible, in particular, I've been having issues with this line: https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/blob/e7c67bd24c454f0b8179032ecbb82284887159c0/src/util/test_util.jl#L208

Should also resolve https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/issues/258

**Proposed changes**

Provide a `jitter` kwarg to the test utils which defaults to whatever the default is in AbstractGPs, but lets the user up it if it's useful for them to do so.

The only alternative I can imagine is forcing the user to pick inputs carefully to ensure that their covariance matrices are always well-conditioned. This is a pain in general.

**Breaking changes**

N/A

edit: ~~will update docstring~~ have updated docstring


theogf (approved):

willtebbutt:
Thanks for the review @theogf . Will merge + tag when CI finishes.

PR #269, opened 2022-03-01
Update recommended audience for API docs

willtebbutt:
The docs previously suggested that they were only for developers. I don't really like warding people off from these docs because they're quite useful. So I'm proposing to change it to suggesting that people read around a little before reading them (I can't imagine anyone getting much out of them without having at least some sense of context).

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -2,8 +2,8 @@
 
 ## Intended Audience
 
-This page is intended for developers.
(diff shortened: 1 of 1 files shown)


theogf (approved):

willtebbutt:
Nightly is taking the lifetime of the universe to finish, so I'm just going to merge.

PR #273, opened 2022-03-15
Parametric Heteroscedastic Example

willtebbutt:
**Summary**

An example that shows off one of the benefits of allowing the user to specify the observation variance.
This is only marginally more involved than a standard regression example.
This model came up when I was chatting to @tpgillam , and it seemed sufficiently simple to implement that I might as well just do it.

**Proposed changes**

N/A

N/A

**Breaking changes**

N/A


examples/parametric-heteroscedastic/script.jl
```diff
@@ -0,0 +1,82 @@
...
+
+# Specify hyperparameters.
+flat_init_params, unflatten = ParameterHandling.value_flatten((
+    s=positive(1.0), l=positive(3.0), σ²=positive(0.1)
+));
+θ_init = unflatten(flat_init_params);
+
+# Build inputs.
+const x = range(0.0, 10.0; length=100)
+const y = rand(Xoshiro(123456), build_gp(θ_init)(x, observation_variance(θ_init, x)))
```
theogf, line 34:
```suggestion
const y = rand(Xoshiro(123456), build_gp(θ_init)(x, observation_variance(θ_init, x)));
```

examples/parametric-heteroscedastic/script.jl
```diff
@@ -0,0 +1,82 @@
...
+
+# Specify objective function.
+function objective(θ)
+    f = build_gp(θ)
+    Σ = observation_variance(θ, x)
+    return -logpdf(f(x, Σ), y)
+end
```
theogf, line 41:
```suggestion
end;
```

examples/parametric-heteroscedastic/script.jl
```diff
@@ -0,0 +1,82 @@
...
+
+# Optimise the hyperparameters. They've been initialised to the correct values, so
```
theogf, line 43: Maybe it's worth to shift them by 1 in log-space or something like this

examples/parametric-heteroscedastic/script.jl
```diff
@@ -0,0 +1,82 @@
...
+    palette=(color=Set1_4,),
+    patchcolor=(Set1_4[2], 0.2),
+    Axis=(limits=((0, 10), nothing),),
+)
+#! format: on
+
+let
+    fig = Figure()
+    ax = Axis(fig[1, 1])
+    plot!(ax, x, f_post(x, Σ_obs_final); bandscale=3)
```
theogf, line 77: Maybe have some labels so we know what we are looking at on the plot

examples/parametric-heteroscedastic/script.jl
willtebbutt: Have resolved locally.

examples/parametric-heteroscedastic/script.jl
willtebbutt: Have resolved locally.

examples/parametric-heteroscedastic/script.jl
willtebbutt: Yup, that would be helpful. Am doing.

willtebbutt:
Hmm @theogf for some reason the docs build isn't updating properly, so I can't look at the changes directly. I'm confident that I've addressed your concerns though -- are you happy to approve, or would you prefer to figure out why we're not getting the right preview?

theogf:
@willtebbutt Now there is always a delay for gh-pages because of this action: https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/actions/workflows/pages/pages-build-deployment

willtebbutt:
@theogf seems to be doing something right now.

theogf:
Sorry  to be picky but could you put a much higher alpha (0.4 or 0.5) for the posterior + noise, I could not see anything at first 
image

theogf (approved):

willtebbutt:
Np. Have upped it. Will merge when CI passes, subject to the plot not needing another tweak.

PR #288, opened 2022-03-17
Improve `diag_At_A` and `diag_At_B` for vectors

sharanry:
```julia
julia> using BenchmarkTools

julia> x = rand(5000); y = rand(5000);

# before
julia> @btime vec(sum(abs2.(x); dims=1));
  2.800 μs (6 allocations: 39.23 KiB)

julia> @btime vec(sum(x .* y; dims=1));
  2.391 μs (5 allocations: 39.23 KiB)

# after
julia> @btime [x'x];
  515.724 ns (3 allocations: 96 bytes)

julia> @btime [x'y];
  813.826 ns (3 allocations: 96 bytes)
```

### DIFF
--- a/src/util/common_covmat_ops.jl
+++ b/src/util/common_covmat_ops.jl
@@ -60,7 +60,8 @@ Xt_invA_Y(X::AbstractVecOrMat, A::Cholesky, Y::AbstractVecOrMat) = (A.U' \ X)' *
 
 At_A(A::AbstractVecOrMat) = A'A
 
-diag_At_A(A::AbstractVecOrMat) = vec(sum(abs2.(A); dims=1))
+diag_At_A(A::AbstractMatrix) = vec(sum(abs2.(A); dims=1))
+diag_At_A(x::AbstractVector) = [x'x]
 
 tr_At_A(A::AbstractVecOrMat) = sum(abs2, A)
 
@@ -72,6 +73,7 @@ function diag_At_B(A::AbstractVecOrMat, B::AbstractVecOrMat)
     )
     return vec(sum(A .* B; dims=1))
 end
+diag_At_B(x::AbstractVector, y::AbstractVector) = [x'y]
 
 diag_Xt_A_X(A::Cholesky, X::AbstractVecOrMat) = diag_At_A(A.U * X)
 


src/util/common_covmat_ops.jl
```diff
@@ -60,7 +60,8 @@ Xt_invA_Y(X::AbstractVecOrMat, A::Cholesky, Y::AbstractVecOrMat) = (A.U' \ X)' *
 
 At_A(A::AbstractVecOrMat) = A'A
 
-diag_At_A(A::AbstractVecOrMat) = vec(sum(abs2.(A); dims=1))
+diag_At_A(A::AbstractMatrix) = vec(sum(abs2.(A); dims=1))
+diag_At_A(x::AbstractVector) = [x'x]
```
devmotion, line 64: I think this should not return a vector since that's not what you would get from `diag(x' * x)`. It seems it's used by `diag_Xt_A_X` but there it's the same, I think it should return a scalar as well since it's just an inner product.

devmotion: Maybe we could even just define generic defaults that correspond to what you're supposed ro get. Eg.,
```julia
diag_At_A(A) = diag(transpose(A) * A)
```

willtebbutt: The current behaviour is to return a 1-element vector, so I would rather keep that behaviour. I don't particularly disagree with your point, I just don't want to make a breaking change in the absence of a compelling reason for doing so (i.e. one of the form "I really need this to return a scalar for my use case").

devmotion: I see your point. The main reason would be that I would expect these functions to be more performant implementations of eg `diag(x' * x)` but not that they return different results/types.

devmotion: It would also be interesting to check if there's a performance difference between `x'*x` and `dot(x, x)` or `sum(abs2, x)` if `x` is a vector.

devmotion: Eg, `sum` is optimized for FillArrays IIRC but I'm not sure about the alternatives.

sharanry:
```julia
julia> x = rand(5000);

julia> @btime vec(sum(abs2.(x); dims=1));
  3.931 μs (5 allocations: 39.20 KiB)

julia> @btime x'x;
  431.201 ns (2 allocations: 32 bytes)

julia> @btime dot(x, x);
  415.221 ns (1 allocation: 16 bytes)

julia> @btime sum(abs2, x);
  778.292 ns (1 allocation: 16 bytes)
```

`dot` seems marginally better.

sharanry: My issue with generic defaults here is that I would want such fundamental operations to error out instead of being secretly non-performant.

devmotion: Can you rerun the benchmarks with interpolated variables? Using global variables can lead to incorrect results. I also think there shouldn't be any allocations.

Ie. can you run
```julia
julia> f(x) = x'*x;

julia> fdot(x) = dot(x, x);

julia> fsum(x) = sum(abs2, x);

julia> @btime f($x);

julia> @btime fdot($x);

julia> @btime fsum($x);
```

sharanry: Yeah you are right. No real difference between `f` and `fdot`
```julia
julia> f_(x) = x'*x;

julia> fdot(x) = dot(x, x);

julia> fsum(x) = sum(abs2, x);

julia> @btime f_($x);
  399.595 ns (0 allocations: 0 bytes)

julia> @btime fdot($x);
  399.545 ns (0 allocations: 0 bytes)

julia> @btime fsum($x);
  469.332 ns (0 allocations: 0 bytes)
```

sharanry: But huge improvement over current master:

```julia
julia> fvec(x) = vec(sum(abs2.(x); dims=1))
fvec (generic function with 1 method)

julia> @btime fvec(x);
  3.499 μs (3 allocations: 39.17 KiB)
```

sharanry: Still a a big improvement over current master:
```julia
julia> fvec(x) = vec(sum(abs2.(x); dims=1))
fvec (generic function with 1 method)

julia> @btime fvec($x);
  3.511 μs (3 allocations: 39.17 KiB)
```

src/util/common_covmat_ops.jl
```diff
@@ -60,7 +60,8 @@ Xt_invA_Y(X::AbstractVecOrMat, A::Cholesky, Y::AbstractVecOrMat) = (A.U' \ X)' *
 
 At_A(A::AbstractVecOrMat) = A'A
 
-diag_At_A(A::AbstractVecOrMat) = vec(sum(abs2.(A); dims=1))
+diag_At_A(A::AbstractMatrix) = vec(sum(abs2.(A); dims=1))
```
st--, line 63: While we're improving this,
```suggestion
diag_At_A(A::AbstractMatrix) = vec(sum(abs2, A); dims=1))
```
is also about a factor 4-5x faster (40-50µs vs 230-260µs on my laptop for a 500x500 matrix), and avoids allocations.

src/util/common_covmat_ops.jl
```diff
@@ -60,7 +60,8 @@ Xt_invA_Y(X::AbstractVecOrMat, A::Cholesky, Y::AbstractVecOrMat) = (A.U' \ X)' *
...
+diag_At_A(x::AbstractVector) = [dot(x, x)]
```
st--, line 64: At least on my laptop, the above definition of
At least on my laptop,
```julia
diag_AtA(x::Vector) = vec(sum(abs2, x; dims=1))
```
is still a factor 2 faster than the `[dot(x, x)]` version...

st--: so maybe better to just keep it the same for Matrix & Vector, but change it from broadcasting `sum(abs2.(A); dims=1)` to `sum(abs2, A; dims=1)`?

src/util/common_covmat_ops.jl
devmotion: In my benchmarks `sum` is much faster:
```julia

julia> using BenchmarkTools, LinearAlgebra

julia> x = rand(50_000);

julia> f(x) = x'*x;

julia> fdot(x) = dot(x, x);

julia> fsum(x) = sum(abs2, x);

julia> @btime f($x);
  8.407 μs (0 allocations: 0 bytes)

julia> @btime fdot($x);
  8.417 μs (0 allocations: 0 bytes)

julia> @btime fsum($x);
  3.429 μs (0 allocations: 0 bytes)
``

File: src/util/common_covmat_ops.jl (hunk 3 above)
Comment (sharanry) [reply src/util/common_covmat_ops.jl:64]:
```julia
julia> f_(x) = x'*x;

julia> fdot(x) = dot(x, x);

julia> fsum(x) = sum(abs2, x);

julia> fvec(x) = vec(sum(abs2.(x); dims=1))
fvec (generic function with 1 method)

julia> fvec2(x) = vec(sum(abs2, x; dims=1))
fvec2 (generic function with 1 method)

julia> @btime f_($x);
  400.340 ns (0 allocations: 0 bytes)

julia> @btime fdot($x);
  400.045 ns (0 allocations: 0 bytes)

julia> @btime fsum($x);
  479.533 ns (0 allocations: 0 bytes)

julia> @btime fvec($x);
  1.639 μs (3 allocations: 39.17 KiB)

julia> @btime fvec2($x);
  625.135 ns (1 allocation: 64 bytes)
```

I see a 1.5x speed for `dot`.

src/util/common_covmat_ops.jl
devmotion: Also for smaller sizes:
```julia
julia> x = rand(5_000);

julia> @btime f($x);
  608.506 ns (0 allocations: 0 bytes)

julia> @btime fdot($x);
  608.403 ns (0 allocations: 0 bytes)

julia> @btime fsum($x);
  337.972 ns (0 allocations: 0 bytes)
``

File: src/util/common_covmat_ops.jl (hunk 3 above)
Comment (devmotion) [reply src/util/common_covmat_ops.jl:64]: This is surprising, I observed a significant improvement with `sum`: https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/pull/288#discussion_r829079380

Comment (st--) [reply src/util/common_covmat_ops.jl:64]: You're not quite comparing like-for-like: if we want to preserve current behaviour, you would need to define
```julia
julia> f_(x) = [x'*x];
julia> fdot(x) = [dot(x, x)];
julia> fsum(x) = [sum(abs2, x)];
```

src/util/common_covmat_ops.jl
```diff
@@ -72,6 +73,7 @@ function diag_At_B(A::AbstractVecOrMat, B::AbstractVecOrMat)
     )
     return vec(sum(A .* B; dims=1))
 end
+diag_At_B(x::AbstractVector, y::AbstractVector) = [dot(x, y)]
```
st--, line 76: I can't comment on that line directly, but should line 67/68 then change as
```diff
-function diag_At_B(A::AbstractVecOrMat, B::AbstractVecOrMat)
+function diag_At_B(A::AbstractMatrix, B::AbstractMatrix)
```
as well?

src/util/common_covmat_ops.jl
sharanry: Could julia version make a difference?
```
Julia Version 1.7.1
Commit ac5cc99908 (2021-12-22 19:35 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: AMD EPYC 7513 32-Core Processor
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-12.0.1 (ORCJIT, znver3)
Environment:
```

src/util/common_covmat_ops.jl
sharanry: I was trying cover mixed cases like `diag_At_B(A::AbstractVector, B::AbstractMatrix)` and `diag_At_B(A::AbstractMatrix, B::AbstractVector)` within the same function.

src/util/common_covmat_ops.jl
sharanry:
```julia
julia> using BenchmarkTools, LinearAlgebra

julia> x = rand(5000);

julia> f_(x) = [x'*x];

julia> fdot(x) = [dot(x, x)];

julia> fsum(x) = [sum(abs2, x)];

julia> fvec(x) = vec(sum(abs2.(x); dims=1));

julia> fvec2(x) = vec(sum(abs2, x; dims=1));

julia> @btime f_($x);
  424.467 ns (1 allocation: 64 bytes)

julia> @btime fdot($x);
  424.518 ns (1 allocation: 64 bytes)

julia> @btime fsum($x);
  487.738 ns (1 allocation: 64 bytes)

julia> @btime fvec($x);
  1.746 μs (3 allocations: 39.17 KiB)

julia> @btime fvec2($x);
  609.023 ns (1 allocation: 64 bytes)
```

sharanry: M1 Mac:
```julia
julia> @btime f_($x);
  2.315 μs (1 allocation: 64 bytes)

julia> @btime fdot($x);
  2.315 μs (1 allocation: 64 bytes)

julia> @btime fsum($x);
  1.079 μs (1 allocation: 64 bytes)

julia> @btime fvec($x);
  1.573 μs (3 allocations: 39.17 KiB)

julia> @btime fvec2($x);
  1.175 μs (1 allocation: 64 bytes)
```

Since `fsum` seems to be performing better in more environments, I am okay with switching to that.

devmotion (changes requested):
Looks good to me. Can you update the benchmark in the initial post and bump the version number? I think it would also be good to add some tests, e.g. with `@allocated`, to ensure there are no major regressions.

src/util/common_covmat_ops.jl
st--: it only works if `size(A) == size(B)`, so the mixed cases will fail anyways

st-- (approved):
looks good to me, thanks for contributing:) @devmotion please merge if you don't have any further change requests

sharanry:
@devmotion Is this good to merge?
I have :
- Bumped the patch version
- Added allocation tests
- Updated the initial post

PR #290, opened 2022-03-21
assert -> explicit throw

st--:
**Summary**

Replace `@assert` with an explicit `throw()`.

**Follow-up question**

Should we also do the same for the prior consistency checks,
```julia
@assert vfe.fz.f === fx.f
```
could move those into `consistency_check` as well? (NB- `check_consistency` would seem to be a slightly more helpful name, functions-as-verbs, any objections to renaming it? it's meant as an internal function, not exported, so shouldn't break backwards compatibility, or would it?)

### DIFF
--- a/src/sparse_approximations.jl
+++ b/src/sparse_approximations.jl
@@ -283,7 +283,7 @@ end
 function consistency_check(fx, y)
     length(fx) == length(y) || throw(
         DimensionMismatch(
-            "length(fx) = $(length(fx)) and length(y) = $(length(y)) must match"
+            "the dimension of the projected GP (here: $(length(fx))) must equal the number of targets (here: $(length(y)))"
         ),
     )
     return nothing
--- a/src/sparse_approximations.jl
+++ b/src/sparse_approximations.jl
@@ -268,7 +268,12 @@ end
 
 # Factor out computations common to the `elbo` and `dtc`.
 function _compute_intermediates(fx::FiniteGP, y::AbstractVector{<:Real}, fz::FiniteGP)
-    consistency_check(fx, y)
+    length(fx) == length(y) || throw(
+        DimensionMismatch(
+            "the dimension of the projected GP (here: $(length(fx))) must equal the number of targets (here: $(length(y)))",
+        ),
+    )
+
     chol_Σy = _cholesky(fx.Σy)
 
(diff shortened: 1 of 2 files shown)


src/sparse_approximations.jl
```diff
@@ -281,7 +281,12 @@ function _compute_intermediates(fx::FiniteGP, y::AbstractVector{<:Real}, fz::Fin
 end
 
 function consistency_check(fx, y)
-    @assert length(fx) == length(y)
+    length(fx) == length(y) || throw(
+        DimensionMismatch(
+            "length(fx) = $(length(fx)) and length(y) = $(length(y)) must match"
```
devmotion, line 286: Could we use a clearer error message? Probably users won't know what `fx` and `y` are.

devmotion (changes requested):

src/sparse_approximations.jl
st--: it's the same variable names as the arguments in the docstrings of `elbo` and `dtc`, but how about
```suggestion
            "The FiniteGP's length(fx) = $(length(fx)) and the targets's length(y) = $(length(y)) must match"
```
?

devmotion: That's not guaranteed though (we could change them, use the check in other fumctions etc.), and even if the arguments the same names users might not be aware of these names. Therefore I don't think the variable names are helpful at all and suggest removing them completely. Maybe just use something like
```julia
"the dimension of the projected GP ($(length(fx))) and the number of targets ($(length(y))) must be equal"
```

st--: If you've got such concrete ideas/thoughts on it, why don't you just go and make it as a code suggestion?

devmotion: Actually I wanted to make one but I didn't have time to deal with the inconveniences of the Github app on my phone. It doesn't allow me to add suggestions in follow-up comments via the interface but requires manually copying and modifying the code. I can add a suggestion this evening/night when I have access to my computer.

st--: fair enough. how about:
```suggestion
            "the dimension of the projected GP (here: $(length(fx))) must equal the number of targets (here: $(length(y)))"
```

PR #303, opened 2022-03-30
bump compat: 0-intro-1d

st--:
Supersedes, #206, #270, #299 and #300.

See #206 for some (now presumably outdated/resolved) discussion around LogDensityProblems.


willtebbutt (approved):
Approved subject to CI passing

PR #313, opened 2022-04-06
reactivate AD tests: mean functions

### LINKED ISSUE #311: AD tests
AbstractGPs is missing AD tests that would uncover when we make changes that work fine in forward mode, but break some autodiff package.

st--:
Turns out there were a bunch of AD tests in the code. They were just commented out (and a bit broken). This is the work-in-progress attempt to reactivate them, and thereby resolve #311. Help welcome!


src/mean_function.jl
```diff
@@ -10,7 +12,7 @@ struct ZeroMean{T<:Real} <: MeanFunction end
 """
 This is an AbstractGPs-internal workaround for AD issues; ideally we would just extend Base.map
 """
-_map_meanfunction(::ZeroMean{T}, x::AbstractVector) where {T} = Zeros{T}(length(x))
```
st--, line 13: turns out this didn't work with AD after all :disappointed:

src/mean_function.jl
```diff
@@ -1,5 +1,7 @@
 abstract type MeanFunction end
 
+# (m::MeanFunction)(x::AbstractVector) = _map_meanfunction(m, x)
```
st--, line 3: should we define something like this ?

test/mean_function.jl
```diff
@@ -1,47 +1,53 @@
...
-        @test iszero(Δx)
+            # Manually verify the ChainRule. Really, this should employ FiniteDifferences, but
+            # currently ChainRulesTestUtils isn't up to handling this, so this will have to do
+            # for now.
+            y, pb = rrule(AbstractGPs._map_meanfunction, m, x)
+            @test y == AbstractGPs._map_meanfunction(m, x)
+            Δmap, Δf, Δx = pb(randn(rng, N))
+            @test iszero(Δmap)
+            @test iszero(Δf)
+            @test iszero(Δx)
```
st--, line 22: is this actually needed, given that we call `differentiable_mean_function_tests` on it anyways?

test/mean_function.jl
```diff
@@ -1,47 +1,53 @@
...
+
         foo_mean = x -> sum(abs2, x)
-        f = CustomMean(foo_mean)
+        m = CustomMean(foo_mean)
 
-        @test AbstractGPs._map_meanfunction(f, x) == map(foo_mean, x)
-        # differentiable_mean_function_tests(f, randn(rng, N), x)
+        for x in [x1, xD, xD′]
+            @test AbstractGPs._map_meanfunction(m, x) == map(foo_mean, x)
+            differentiable_mean_function_tests(m, randn(rng, N), x)
```
st--, line 50: Should we just call
```suggestion
            differentiable_mean_function_tests(rng, m, x)
```
instead (and remove the y = ...), or otherwise remove that (currently unused) method definition of differentiable_mean_function_tests?

test/Project.toml
```diff
@@ -1,4 +1,5 @@
 [deps]
+AbstractGPs = "99985d1d-32ba-4be9-9821-2ec096f28918"
```
devmotion, line 2: The tested package is usually not part of test/Project.toml as Pkg adds it automatically (https://pkgdocs.julialang.org/v1/creating-packages/#Test-specific-dependencies-in-Julia-1.2-and-above) and otherwise one has to add and update compat entries:
```suggestion
```

st--: Ok, I added it because I wanted to be able to run some of the tests locally, and when I used julia --project=test it complained about AbstractGPs not being in the project

devmotion: Yeah, that's not an intended workflow and not officially supported. Tests are supposed to be run with `Pkg.test`.

st--: so what's the intended workflow for "I don't want to run all the tests because that takes a really long time, I just want to run the tests that I'm currently working on in this one file"?

st--: (who says what's intended, officially supported, and supposed to be done? where would I find out about that?)

devmotion: Eg., comment the others out, specify environment variables for specific test sections (if supported), or use TestEnv.

devmotion: With the test setup in eg KernelFunctions (and AbstractGPs?) with all imports and utilities in runtests.jl one has to run runtests.jl anyway, it seems, or load the packages manually. In general, also the first two options might be a bit misleading if eg other tests mutate the RNG.

theogf: @st-- I use TestEnv.jl  (loaded at startup). In the working repo:
```
TestEnv.activate()
```
and then all tests modules are loaded as well as the current repo.
But I agree it's a mess

st--: Ok, TestEnv seems to work, so I've reverted this change. How would I have been able to find this out by myself? Is this something we should/could document somewhere ?

st--: I like a lot of things about Julia (e.g. the Pkg manager is so much better than in python land). But I have to say I miss `pytest`. It makes it so easy to select which subset of tests you want to run. E.g. just the slow ones, or just the fast ones, or the ones in these files only...

I think it'd help if instead of stuffing all imports into runtests.jl, having them at the top of each individual test file. Then it'd be easy to just `include()` one of them.

devmotion: I guess one learns about such things by eg asking, googling, or attending JuliaCon. IIRC a while ago TestEnv was also discussed in the Turing slack (probably during or after JuliaCon).

I don't think we should document anything here. It's not a common task, it's not JuliaGP specific, and there's not a single preferred approach in the Julia ecosystem. In particular we don't care about how people test their PRs locally, if they run all tests or only parts of it (this can be tricky even with separate files +TestEnv, see below) and if they use TestEnv or not.

One nice thing about tests is that it is very flexible since it's just one long script, possibly with `include`s. One can split it up or just use a single file, one can use multiple test sets or not, one can run some tests conditionally on environment variables... But of course the flexibility comes at a cost, eg. when trying to run a subset of tests only.

In general, running subsets of tests seems most convenient if they are 1) separated in different files and/or with different switches such as environment variables (used eg in SciML but also the Turing ecosystem and even AbstractGPs) and 2) put into separate modules, eg with SafeTestsets, with explicit imports, test utilities, and generally avoiding any other leakage eg from mutating the RNG. Without 2) running tests separately may fail and, if all are passing, does not guarantee that running all tests together is successful.

src/mean_function.jl
devmotion: What would be the motivation? Generally, I don't like implicit mapping or broadcasting. IIRC the function only exists to work around Zygote AD issues.

In any case, IMO it does not belong kn this PR.

src/mean_function.jl
```diff
@@ -1,5 +1,7 @@
...
+
```
devmotion, line 4:
```suggestion
```

src/mean_function.jl
devmotion: Did you check if one can just fix the rrule below?

src/mean_function.jl
```diff
@@ -28,7 +30,7 @@ struct ConstMean{T<:Real} <: MeanFunction
     c::T
 end
 
-_map_meanfunction(m::ConstMean, x::AbstractVector) = Fill(m.c, length(x))
+_map_meanfunction(m::ConstMean, x::AbstractVector) = fill(m.c, length(x))
```
devmotion, line 33: Same here, what's the actual problem?

test/mean_function.jl
```diff
@@ -1,47 +1,53 @@
 @testset "mean_functions" begin
     @testset "ZeroMean" begin
-        P = 3
-        Q = 2
-        D = 4
-        # X = ColVecs(randn(rng, D, P))
-        x = randn(P)
-        x̄ = randn(P)
-        f = ZeroMean{Float64}()
+        rng, N, D = MersenneTwister(123456), 5, 3
```
devmotion, line 3: IMO it's more readable to have separate assignments (wonder if it's even discussed by BlueStyle):

```suggestion
        rng = MersenneTwister(123456)
        N = 5
        D = 3
```

test/mean_function.jl
```diff
@@ -1,47 +1,53 @@
...
 
-        for x in [x]
-            @test AbstractGPs._map_meanfunction(f, x) == zeros(size(x))
-            # differentiable_mean_function_tests(f, randn(rng, P), x)
-        end
+        m = ZeroMean{Float64}()
+
+        for x in [x1, xD, xD′]
+            @test AbstractGPs._map_meanfunction(m, x) == zeros(size(x))
+            differentiable_mean_function_tests(m, randn(rng, N), x)
```
devmotion, line 12: In general, it seems there are many more changes than needed here? This makes it a bit difficult to review. It seems sufficient to uncomment this line and test ColVecs and RowVecs as well.

test/mean_function.jl
devmotion, line 22: Same here, it seems only variables are renamed?

test/mean_function.jl
```diff
@@ -1,47 +1,53 @@
...
+            @test iszero(Δmap)
+            @test iszero(Δf)
+            @test iszero(Δx)
+        end
     end
     @testset "ConstMean" begin
-        rng, D, N = MersenneTwister(123456), 5, 3
-        # X = ColVecs(randn(rng, D, N))
-        x = randn(rng, N)
+        rng, N, D = MersenneTwister(123456), 5, 3
```
devmotion, line 26: Why do you swap N and D? Seems not necessary.

test/mean_function.jl
```diff
@@ -1,47 +1,53 @@
...
-        rng, N, D = MersenneTwister(123456), 11, 2
-        x = randn(rng, N)
+        rng, N, D = MersenneTwister(123456), 5, 3
+        x1 = randn(rng, N)
+        xD = ColVecs(randn(rng, D, N))
+        xD′ = RowVecs(randn(rng, N, D))
+
         foo_mean = x -> sum(abs2, x)
-        f = CustomMean(foo_mean)
+        m = CustomMean(foo_mean)
```
devmotion, line 46: Also unrelated it seems:
```suggestion
        f = CustomMean(foo_mean)
```

test/test_util.jl
```diff
@@ -27,6 +27,10 @@ end
 Base.zero(d::Dict) = Dict([(key, zero(val)) for (key, val) in d])
 Base.zero(x::Array) = zero.(x)
 
+# TODO should move into KernelFunctions.jl
+Base.zero(x::ColVecs) = ColVecs(zero(x.X))
+Base.zero(x::RowVecs) = RowVecs(zero(x.X))
+
```
devmotion, line 33: Yeah this should really not exist here.

```suggestion
```

devmotion (dismissed):

test/mean_function.jl
st--: so that it's consistent. it was N, D in some places and D, N in others. N, D seemed more natural to me so I changed the D, N to that instead of the other way around

test/mean_function.jl
st--: again, it was `m` in some places and `f` in others, and I wanted to make it consistent. `m` seemed to make more sense to me (it's not a function, can't be called as it is at the moment)

test/test_util.jl
st--: :+1: I'll open a PR in KernelFunctions

st--: I'll update it here once https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/444 is merged

src/mean_function.jl
st--: that the AD tests failed

src/mean_function.jl
st--: I don't know how you would fix it.

src/mean_function.jl
st--: if you know how to write the custom rrule to fix it with keeping the `Fill`, please suggest!!

src/mean_function.jl
st--: no, it doesn't belong in here, I just thought while people are thinking about mean functions we can consider it. will remove it before merging.

test/test_util.jl
st--: Though note that this is just for the tests. Above, it also defines
```julia
Base.zero(d::Dict) = Dict([(key, zero(val)) for (key, val) in d])
Base.zero(x::Array) = zero.(x)
```
(The zero(::Array) seems to actually be implemented in Base, so not sure why that's there)

st--:
@devmotion you requested changes, but it's not clear to me what changes you're actually requesting, could you please clarify?
regarding the definition of zero(::ColVecs), see my reply https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/pull/313#discussion_r844950873 though I also opened https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/444

src/mean_function.jl
devmotion: No, I don't because I don't know what's th e error (which was the point of my comment above). Don't you have a stacktrace?

devmotion:
I would like my comments to be addressed. Ie, no new commented out methods, no unrelated and unneeded changes, no new type piracy (doesn't matter if it's for tests or in the package), and understanding what causes the AD issues before reverting the FillArray changes.

src/mean_function.jl
st--:
```
ZeroMean: Error During Test at /u/10/johnt1/unix/JuliaCode/AbstractGPs.jl/test/mean_function.jl:2
  Got exception outside of a @test
  DimensionMismatch("second dimension of A, 1, does not match length of x, 5")
  Stacktrace:
    [1] gemv!(y::Vector{Float64}, tA::Char, A::Matrix{Float64}, x::Vector{Float64}, α::Bool, β::Bool)
      @ LinearAlgebra ~/Software/julia-1.7.1/share/julia/stdlib/v1.7/LinearAlgebra/src/matmul.jl:530
    [2] mul!
      @ ~/Software/julia-1.7.1/share/julia/stdlib/v1.7/LinearAlgebra/src/matmul.jl:97 [inlined]
    [3] mul!
      @ ~/Software/julia-1.7.1/share/julia/stdlib/v1.7/LinearAlgebra/src/matmul.jl:275 [inlined]
    [4] *(transA::Transpose{Float64, Matrix{Float64}}, x::Vector{Float64})
      @ LinearAlgebra ~/Software/julia-1.7.1/share/julia/stdlib/v1.7/LinearAlgebra/src/matmul.jl:87
    [5] _j′vp(fdm::FiniteDifferences.AdaptedFiniteDifferenceMethod{5, 1, FiniteDifferences.UnadaptedFiniteDifferenceMethod{7, 5}}, f::Function, ȳ::Vector{Float64}, x::Vector{Float64})
      @ FiniteDifferences ~/.julia/packages/FiniteDifferences/VpgIT/src/grad.jl:80
    [6] j′vp
      @ ~/.julia/packages/FiniteDifferences/VpgIT/src/grad.jl:73 [inlined]
    [7] adjoint_test(f::Function, ȳ::Vector{Float64}, x::Vector{Float64}; rtol::Float64, atol::Float64, fdm::FiniteDifferences.AdaptedFiniteDifferenceMethod{5, 1, FiniteDifferences.UnadaptedFiniteDifferenceMethod{7, 5}}, print_results::Bool)
      @ Main ~/JuliaCode/AbstractGPs.jl/test/test_util.jl:64
    [8] differentiable_mean_function_tests(m::ZeroMean{Float64}, ȳ::Vector{Float64}, x::Vector{Float64}; rtol::Float64, atol::Float64)
      @ Main ~/JuliaCode/AbstractGPs.jl/test/test_util.jl:102
    [9] differentiable_mean_function_tests(m::ZeroMean{Float64}, ȳ::Vector{Float64}, x::Vector{Float64})
      @ Main ~/JuliaCode/AbstractGPs.jl/test/test_util.jl:98
   [10] macro expansion
      @ ~/JuliaCode/AbstractGPs.jl/test/mean_function.jl:12 [inlined]
   [11] macro expansion
      @ ~/Software/julia-1.7.1/share/julia/stdlib/v1.7/Test/src/Test.jl:1283 [inlined]
   [12] macro expansion
      @ ~/JuliaCode/AbstractGPs.jl/test/mean_function.jl:3 [inlined]
   [13] macro expansion
      @ ~/Software/julia-1.7.1/share/julia/stdlib/v1.7/Test/src/Test.jl:1283 [inlined]
   [14] top-level scope
      @ ~/JuliaCode/AbstractGPs.jl/test/mean_function.jl:2
   [15] include(fname::String)
      @ Base.MainInclude ./client.jl:451
   [16] top-level scope
      @ REPL[22]:1
   [17] eval
      @ ./boot.jl:373 [inlined]
   [18] eval_user_input(ast::Any, backend::REPL.REPLBackend)
      @ REPL ~/Software/julia-1.7.1/share/julia/stdlib/v1.7/REPL/src/REPL.jl:150
   [19] repl_backend_loop(backend::REPL.REPLBackend)
      @ REPL ~/Software/julia-1.7.1/share/julia/stdlib/v1.7/REPL/src/REPL.jl:244
   [20] start_repl_backend(backend::REPL.REPLBackend, consumer::Any)
      @ REPL ~/Software/julia-1.7.1/share/julia/stdlib/v1.7/REPL/src/REPL.jl:229
   [21] run_repl(repl::REPL.AbstractREPL, consumer::Any; backend_on_current_task::Bool)
      @ REPL ~/Software/julia-1.7.1/share/julia/stdlib/v1.7/REPL/src/REPL.jl:362
   [22] run_repl(repl::REPL.AbstractREPL, consumer::Any)
      @ REPL ~/Software/julia-1.7.1/share/julia/stdlib/v1.7/REPL/src/REPL.jl:349
   [23] (::Base.var"#930#932"{Bool, Bool, Bool})(REPL::Module)
      @ Base ./client.jl:394
   [24] #invokelatest#2
      @ ./essentials.jl:716 [inlined]
   [25] invokelatest
      @ ./essentials.jl:714 [inlined]
   [26] run_main_repl(interactive::Bool, quiet::Bool, banner::Bool, history_file::Bool, color_set::Bool)
      @ Base ./client.jl:379
   [27] exec_options(opts::Base.JLOptions)
      @ Base ./client.jl:309
   [28] _start()
      @ Base ./client.jl:495
```

src/mean_function.jl
st--:
```
ConstMean: Error During Test at /u/10/johnt1/unix/JuliaCode/AbstractGPs.jl/test/mean_function.jl:25
  Got exception outside of a @test
  DimensionMismatch("second dimension of A, 2, does not match length of x, 5")
  Stacktrace:
    [1] gemv!(y::Vector{Float64}, tA::Char, A::Matrix{Float64}, x::Vector{Float64}, α::Bool, β::Bool)
      @ LinearAlgebra ~/Software/julia-1.7.1/share/julia/stdlib/v1.7/LinearAlgebra/src/matmul.jl:530
    [2] mul!
      @ ~/Software/julia-1.7.1/share/julia/stdlib/v1.7/LinearAlgebra/src/matmul.jl:97 [inlined]
    [3] mul!
      @ ~/Software/julia-1.7.1/share/julia/stdlib/v1.7/LinearAlgebra/src/matmul.jl:275 [inlined]
    [4] *(transA::Transpose{Float64, Matrix{Float64}}, x::Vector{Float64})
      @ LinearAlgebra ~/Software/julia-1.7.1/share/julia/stdlib/v1.7/LinearAlgebra/src/matmul.jl:87
    [5] _j′vp(fdm::FiniteDifferences.AdaptedFiniteDifferenceMethod{5, 1, FiniteDifferences.UnadaptedFiniteDifferenceMethod{7, 5}}, f::Function, ȳ::Vector{Float64}, x::Vector{Float64})
      @ FiniteDifferences ~/.julia/packages/FiniteDifferences/VpgIT/src/grad.jl:80
    [6] j′vp
      @ ~/.julia/packages/FiniteDifferences/VpgIT/src/grad.jl:73 [inlined]
    [7] adjoint_test(f::Function, ȳ::Vector{Float64}, x::Vector{Float64}; rtol::Float64, atol::Float64, fdm::FiniteDifferences.AdaptedFiniteDifferenceMethod{5, 1, FiniteDifferences.UnadaptedFiniteDifferenceMethod{7, 5}}, print_results::Bool)
      @ Main ~/JuliaCode/AbstractGPs.jl/test/test_util.jl:64
    [8] #differentiable_mean_function_tests#14
      @ ~/JuliaCode/AbstractGPs.jl/test/test_util.jl:102 [inlined]
    [9] differentiable_mean_function_tests(m::ConstMean{Float64}, ȳ::Vector{Float64}, x::Vector{Float64})
      @ Main ~/JuliaCode/AbstractGPs.jl/test/test_util.jl:98
   [10] macro expansion
      @ ~/JuliaCode/AbstractGPs.jl/test/mean_function.jl:36 [inlined]
   [11] macro expansion
      @ ~/Software/julia-1.7.1/share/julia/stdlib/v1.7/Test/src/Test.jl:1283 [inlined]
   [12] macro expansion
      @ ~/JuliaCode/AbstractGPs.jl/test/mean_function.jl:26 [inlined]
   [13] macro expansion
      @ ~/Software/julia-1.7.1/share/julia/stdlib/v1.7/Test/src/Test.jl:1283 [inlined]
   [14] top-level scope
      @ ~/JuliaCode/AbstractGPs.jl/test/mean_function.jl:2
   [15] include(fname::String)
      @ Base.MainInclude ./client.jl:451
   [16] top-level scope
      @ REPL[22]:1
   [17] eval
      @ ./boot.jl:373 [inlined]
   [18] eval_user_input(ast::Any, backend::REPL.REPLBackend)
      @ REPL ~/Software/julia-1.7.1/share/julia/stdlib/v1.7/REPL/src/REPL.jl:150
   [19] repl_backend_loop(backend::REPL.REPLBackend)
      @ REPL ~/Software/julia-1.7.1/share/julia/stdlib/v1.7/REPL/src/REPL.jl:244
   [20] start_repl_backend(backend::REPL.REPLBackend, consumer::Any)
      @ REPL ~/Software/julia-1.7.1/share/julia/stdlib/v1.7/REPL/src/REPL.jl:229
   [21] run_repl(repl::REPL.AbstractREPL, consumer::Any; backend_on_current_task::Bool)
      @ REPL ~/Software/julia-1.7.1/share/julia/stdlib/v1.7/REPL/src/REPL.jl:362
   [22] run_repl(repl::REPL.AbstractREPL, consumer::Any)
      @ REPL ~/Software/julia-1.7.1/share/julia/stdlib/v1.7/REPL/src/REPL.jl:349
   [23] (::Base.var"#930#932"{Bool, Bool, Bool})(REPL::Module)
      @ Base ./client.jl:394
   [24] #invokelatest#2
      @ ./essentials.jl:716 [inlined]
   [25] invokelatest
      @ ./essentials.jl:714 [inlined]
   [26] run_main_repl(interactive::Bool, quiet::Bool, banner::Bool, history_file::Bool, color_set::Bool)
      @ Base ./client.jl:379
   [27] exec_options(opts::Base.JLOptions)
      @ Base ./client.jl:309
   [28] _start()
      @ Base ./client.jl:495
```

Can you make any sense of this ?

st--:
> no new commented out methods,

I already said I was going to remove it before merging, just left it there while we were still discussing it so others could consider it.

> no unrelated and unneeded changes,

I'm assuming you're referring to cleaning up the tests & making them more consistent (I was working towards being able to unify them further, instead of duplicating [test] code)? I've separated it out into https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/pull/315/files, and temporarily changed the PR target so that those changes don't show up in here.

> no new type piracy (doesn't matter if it's for tests or in the package),

I already opened https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/444 for this. I've now added tests there, and I'll of course update this PR once KernelFunctions has a new release.

> and understanding what causes the AD issues before reverting the FillArray changes (since possibly one can fix them in a better way).

I've posted the stack traces, if you can make some sense of them, please enlighten me! But if it's not straightforward to fix (and at least to me it doesn't seem straightforward), I would think it makes more sense to add AD tests & get them to pass (as otherwise the package is currently broken), and then leave fixing the FillArray changes for later.

st--:
@devmotion I believe that addressed all your concerns

devmotion:
> I've posted the stack traces, if you can make some sense of them, please enlighten me! But if it's not straightforward to fix (and at least to me it doesn't seem straightforward), I would think it makes more sense to add AD tests & get them to pass (as otherwise the package is currently broken), and then leave fixing the FillArray changes for later.

The stracktraces aren't too bad, are they? It's obviously a problem with the calls of FiniteDifferences in https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/blob/598cd63c71ea7dbde451403304df4d41ebe89909/test/test_util.jl#L54-L68. Just using ChainRulesTestUtils with its improved logic in https://github.com/JuliaDiff/ChainRulesTestUtils.jl/blob/main/src/finite_difference_calls.jl (also e.g. for `NoTangent`s etc.) might be sufficient. `rrule`s should be tested with ChainRulesTestUitls anyway instead of some custom Zygote-based tests.

st--:
> The stracktraces aren't too bad, are they?

I'm glad for you that you find parsing them so easy. If I had found it as easy to identify the issue & how to fix it, believe me, I would have just done that. I've already spent more time than I should have on trying to make the situation a bit better. If you want _me_ to do it, all I can do at this point is propose this change to go back to fill() and zeros() and leave the "change to better ways of doing it & revert the FillArrays" to an open issue someone else can pick up at a later date.

test/mean_function.jl
```diff
@@ -10,7 +10,7 @@
 
         for x in [x1, xD, xD′]
             @test AbstractGPs._map_meanfunction(m, x) == zeros(N)
-            #differentiable_mean_function_tests(m, randn(rng, N), x)
+            differentiable_mean_function_tests(m, randn(rng, N), x)
 
             # Manually verify the ChainRule. Really, this should employ FiniteDifferences, but
             # currently ChainRulesTestUtils isn't up to handling this, so this will have to do
```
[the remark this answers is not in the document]
st--: (sometimes it's so annoying that you can't comment on unchanged code more than two lines or so away...)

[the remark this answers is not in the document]
willtebbutt: Yeah, can get rid of them now.

devmotion:
I unsubscribed, please find another reviewer.

test/test_util.jl
```diff
@@ -87,34 +88,17 @@ end
...
-    atol=_atol,
+    m::MeanFunction, ȳ::AbstractVector, x::AbstractVector; rtol=_rtol, atol=_atol
 )
     # Run forward tests.
     mean_function_tests(m, x)
 
     # Check adjoint.
     @assert length(ȳ) == length(x)
-    return adjoint_test(x -> ew(m, x), ȳ, x; rtol=rtol, atol=atol)
+    adjoint_test(x -> AbstractGPs._map_meanfunction(m, x), ȳ, x; rtol=rtol, atol=atol)
```
willtebbutt, line 98:
```suggestion
    adjoint_test(x -> collect(AbstractGPs._map_meanfunction(m, x)), ȳ, x; rtol=rtol, atol=atol)
```
should address the issues discussed in https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/pull/313/files#r844871805

src/mean_function.jl
willtebbutt: Frustratingly, this is much more likely to be an issue with our AD testing infrastructure than the code itself. I've suggested a modification to the code below which _ought_ to resolve the problems.

test/mean_function.jl
willtebbutt: Seems like it would simplify things a bit, so I'm in favour.

src/mean_function.jl
willtebbutt: Should also be resolved via my suggestion below which modifies the test slightly.

test/test_util.jl
willtebbutt: For context: our AD testing infrastructure is really quite reliable when you're working with `Array{Float64, N}`, but is less reliable for other kinds of subtypes of `AbstractArray`. This is obviously really annoying, but we're stuck with it for now.

willtebbutt: For more context: what is probably going on is that `FiniteDifferences.to_vec(::Fill)` returns not quite the right thing. Ideally we would have better debugging tools there, but they've yet to be written for a number of reasons (available person-time, unclear precisely where some of the rough edges are, etc. The usual reasons)

willtebbutt (commented):
I think this broadly looks good -- I've given some thoughts on how we might sort out the AD testing related problems. I would like to re-review once we've sorted out the AD testing issues, but I think it's basically there.

src/mean_function.jl
```diff
@@ -28,7 +28,12 @@ struct ConstMean{T<:Real} <: MeanFunction
...
+
+function ChainRulesCore.rrule(::typeof(_map_meanfunction), m::ConstMean, x::AbstractVector)
+    map_ConstMean_pullback(Δ) = (NoTangent(), Tangent{ConstMean}(; c=sum(Δ)), ZeroTangent())
+    return _map_meanfunction(m, x), map_ConstMean_pullback
+end
```
willtebbutt, line 36: Also, was this introduced to attempt to deal with the AD testing problems?

test/test_util.jl
willtebbutt: It's also true that we're implicitly assuming that a `Vector{Float64}` is an appropriate tangent type for whatever is returned by `mean(f, x)`. This is definitely not true for `Fill`.

Whichever explanation is correct, I _hope_ that the proposed modification resolves the problem. It very slightly weakens the test, but not by much -- I would certainly be convinced that `mean` is being differentiated correctly if it passes.

willtebbutt: (Have just run locally -- seems to resolve the problem)

test/mean_function.jl
st--: which one? remove the (rng, m, x) method, or apply this suggestion?

src/mean_function.jl
```diff
@@ -28,7 +28,12 @@ struct ConstMean{T<:Real} <: MeanFunction
     c::T
 end
 
-_map_meanfunction(m::ConstMean, x::AbstractVector) = Fill(m.c, length(x))
+_map_meanfunction(m::ConstMean, x::AbstractVector) = fill(m.c, length(x))
```
st--, line 31:
```suggestion
_map_meanfunction(m::ConstMean, x::AbstractVector) = Fill(m.c, length(x))
```

src/mean_function.jl
```diff
@@ -10,10 +10,10 @@ struct ZeroMean{T<:Real} <: MeanFunction end
 """
 This is an AbstractGPs-internal workaround for AD issues; ideally we would just extend Base.map
 """
-_map_meanfunction(::ZeroMean{T}, x::AbstractVector) where {T} = Zeros{T}(length(x))
+_map_meanfunction(::ZeroMean{T}, x::AbstractVector) where {T} = zeros(T, length(x))
```
st--, line 13:
```suggestion
 _map_meanfunction(::ZeroMean{T}, x::AbstractVector) where {T} = Zeros{T}(length(x))
```

src/mean_function.jl
st--: yes, that was my attempt at working around the lack of FillArray rrules/projectors by doing it directly on the mean function call... do you think it should get removed again ?

st--:
@willtebbutt thanks, looks like your collect() actually fixed the issues, and now it's all happy even with Zero and Fill. Also, I checked and neither of the two `rrule`s defined in mean_function.jl for ZeroMean and ConstMean, respectively, are actually needed for the AD test to pass. Should we keep (/add) or remove them ?

Re https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/pull/313#discussion_r846658855 please let me know which way around you prefer.

willtebbutt:
> @willtebbutt thanks, looks like your collect() actually fixed the issues, and now it's all happy even with Zero and Fill. Also, I checked and neither of the two rrules defined in mean_function.jl for ZeroMean and ConstMean, respectively, are actually needed for the AD test to pass. Should we keep (/add) or remove them ?

Excellent. Please remove the rules since they're now not needed.

test/mean_function.jl
willtebbutt: I would be inclined to retain the method that just requires a `rng`, since we can now be confident that it will generate an appropriate tangent due to the call to `collect`.

st--:
Will do.

What do you think of https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/pull/313/commits/f1df8b501c0c0c624518b53f7fe0eb89a6839cc2 ?

st--:
> Please remove the rules since they're now not needed.

I'm wondering if they might make it a bit faster though (as they're shortcutting some backprop)?

willtebbutt:
> I'm wondering if they might make it a bit faster though (as they're shortcutting some backprop)?

They really shouldn't make stuff faster in this case -- it's simple enough that Zygote ought to be able to infer this properly. In my view, the additional code complexity probably isn't worth it.

willtebbutt:
> What do you think of https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/commit/f1df8b501c0c0c624518b53f7fe0eb89a6839cc2 ?

I like it. Nice to remove some code duplication.

willtebbutt (approved):
LGTM

PR #314, opened 2022-04-06
reactivate AD tests: FiniteGP

st--:
Follow-up to #313


test/finite_gp_projection.jl
```diff
@@ -177,42 +149,40 @@ end
...
+            atol=1e-8,
+            rtol=1e-8,
+        )
+
+        # Check that the gradient w.r.t. the noise is approximately correct for `f`.
+        σ_ = randn(rng)
+        adjoint_test((σ_, ŷ) -> logpdf(f(x, exp(σ_)), ŷ), l̄, σ_, ŷ)
+        adjoint_test((σ_, Ŷ) -> sum(logpdf(f(x, exp(σ_)), Ŷ)), l̄, σ_, Ŷ)
+
+        # Check that the gradient w.r.t. a scaling of the GP works.
```
st--, line 180: It's not actually implemented in AbstractGPs, so we should remove these (presumably leftover from the initial Stheno port?)

test/sparse_approximations.jl
```diff
@@ -122,11 +122,47 @@ end
...
+        x = collect(range(-3.0, 3.0; length=N))
+        # @testset "dense" begin
+        #     rng = MersenneTwister(123456)
+        #     A = randn(rng, N, N - 2)
+        #     adjoint_test(
+        #         (x, A) -> begin
+        #             f = GP(sin, SqExponentialKernel())
+        #             Σy = _to_psd(A)
+        #             C = cholesky(Σy)
+        #             return AbstractGPs.tr_Cf_invΣy(f(x, Σy), Σy, C)
```
st--, line 145: This isn't actually implemented. I'm wondering if we should just remove these tests entirely. Not sure what the value is in testing gradients of an internal function specifically - as long as the logpdf and mean/cov gradients work fine ... just remove it?

test/sparse_approximations.jl
```diff
@@ -122,11 +122,47 @@ end
     σ² = 1.234
     Σy_Diagonal = Diagonal(Fill(σ², N))
     Σy_ScalMat = ScalMat(N, σ²)
+    Σy_dense = _to_psd(randn(N, N - 2))
     f = GP(SqExponentialKernel())
 
-    for Σy in (Σy_Diagonal, Σy_ScalMat)
+    for Σy in (Σy_Diagonal, Σy_ScalMat) #, Σy_dense)
```
st--, line 128: it's not implemented for dense :see_no_evil:

test/finite_gp_projection.jl
willtebbutt: Yup. These can go.

test/sparse_approximations.jl
willtebbutt: Probably for the best for now. If the observation covariance is dense, I would _imagine_ that performance will generally be worse than exact inference.

test/sparse_approximations.jl
willtebbutt: I'm also happy for these tests to be removed.

willtebbutt (approved):
I'm happy with this, subject to the comments.

PR #323, opened 2022-05-16
Opt out of `rrule` for `mean`

devmotion:
https://github.com/JuliaDiff/ChainRules.jl/pull/615 broke AD since it captures also `mean(::AbstractGP, ::AbstractVector)`. Opting out of the specific rule seems to fix test errors locally.

cc @simsurace


simsurace:
Awesome! This fixes my issue at least.

theogf (approved):

oxinabox:
Can we add AbstractGPs as a reverse dependency test for ChainRules.jl to avoid this in future?

PR #329, opened 2022-07-21
Use non scalar method to compute logdet for cholesky and relax compat for PDMat

sharanry:
This enables the computation of `dtc` on GPU

### DIFF
--- a/src/sparse_approximations.jl
+++ b/src/sparse_approximations.jl
@@ -266,7 +266,7 @@ function dtc(vfe::VFE, fx::FiniteGP, y::AbstractVector{<:Real})
     return _dtc
 end
 
-logdet_(C::Cholesky) = sum(log.(diag(C.factors)))
+logdet_(C::Cholesky) = 2 * sum(log, diag(C.factors))
 # Factor out computations common to the `elbo` and `dtc`.
 function _compute_intermediates(fx::FiniteGP, y::AbstractVector{<:Real}, fz::FiniteGP)
     length(fx) == length(y) || throw(
--- a/src/sparse_approximations.jl
+++ b/src/sparse_approximations.jl
@@ -267,6 +267,7 @@ function dtc(vfe::VFE, fx::FiniteGP, y::AbstractVector{<:Real})
 end
 
 logdet_(C::Cholesky) = 2 * sum(log, diag(C.factors))
+
 # Factor out computations common to the `elbo` and `dtc`.
 function _compute_intermediates(fx::FiniteGP, y::AbstractVector{<:Real}, fz::FiniteGP)
     length(fx) == length(y) || throw(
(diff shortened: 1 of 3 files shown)


sharanry:
Will be creating an issue in https://github.com/JuliaLang/julia

theogf:
Could you add a test to make sure that `logdet_ == logdet` ?

theogf:
Also why the relaxation on the PDMats Compat?

src/sparse_approximations.jl
```diff
@@ -266,6 +266,7 @@ function dtc(vfe::VFE, fx::FiniteGP, y::AbstractVector{<:Real})
     return _dtc
 end
 
+logdet_(C::Cholesky) = sum(log.(diag(C.factors)))
```
theogf, line 269:
```suggestion
logdet_(C::Cholesky) = 2 * sum(log, diag(C.factors)))
```

sharanry:
> Also why the relaxation on the PDMats Compat?

I encounter this without the relaxation
```julia
ERROR: Unsatisfiable requirements detected for package ElasticPDMats [2904ab23]:
 ElasticPDMats [2904ab23] log:
 ├─possible versions are: 0.1.0-0.2.2 or uninstalled
 ├─restricted by compatibility requirements with GaussianProcesses [891a1506] to versions: 0.2.0-0.2.2
 │ └─GaussianProcesses [891a1506] log:
 │   ├─possible versions are: 0.8.0-0.12.4 or uninstalled
 └─restricted by compatibility requirements with PDMats [90014a1f] to versions: uninstalled — no versions left
   └─PDMats [90014a1f] log:
     ├─possible versions are: 0.9.0-0.11.16 or uninstalled
     └─restricted to versions 0.11 by AbstractGPs [99985d1d], leaving only versions 0.11.0-0.11.16
       └─AbstractGPs [99985d1d] log:
         ├─possible versions are: 0.5.14 or uninstalled
```

devmotion:
I don't understand the changes in this PR, can you explain the motivation?

The compat issues seem to be caused by ElasticPDMats/GaussianProcesses/WilliamsGPR. I don't remember the details but there was surely a reason for why we dropped support for PDMata 0.10. So I think it would be better to fix the other problematic packages but not to reintroduce the PDMats 0.10 compat entry.

The change to `_logdet` is also unclear to me as it breaks the package in its current form. I guess you maybe forgot to define `_logdet(x) = logdet(x)` fallback? In any case, I don't think such an internal method for `Cholesky` should be introduced here. The `logdet` definition for `Cholesky` in base (https://github.com/JuliaLang/julia/blob/46a6f2230d9d05f0bb26ca5d241a7316993c10a8/stdlib/LinearAlgebra/src/cholesky.jl#L669) is fine, and I assume GPU support should be added in CUDA.jl or similar packages, probably by defining something like `logdet(::Cholesky{T,CuMatrix{T}})` (see eg also https://github.com/JuliaGPU/CUDA.jl/issues/110).

Issue #333, opened 2022-09-17
 ERROR: MethodError: no method matching Int64(::Irrational{:log2π})


jariji:
I'm trying to learn to use this package. I copied the documentation example of DynamicHMC from https://juliagaussianprocesses.github.io/AbstractGPs.jl/dev/examples/0-intro-1d/

```jl
using AbstractGPs
using Distributions
using StatsFuns

using Plots
default(; legend=:outertopright, size=(700, 400))

using Random
Random.seed!(42)  # setting the seed for reproducibility of this notebook


x = [
    0.8658165855998895,
    0.6661700880180962,
    0.8049218148148531,
    0.7714303440386239,
    0.14790478354654835,
    0.8666105548197428,
    0.007044577166530286,
    0.026331737288148638,
    0.17188596617099916,
    0.8897812990554013,
    0.24323574561119998,
    0.028590102134105955,
]
y = [
    1.5255314337144372,
    3.6434202968230003,
    3.010885733911661,
    3.774442382979625,
    3.3687639483798324,
    1.5506452040608503,
    3.790447985799683,
    3.8689707574953,
    3.4933565751758713,
    1.4284538820635841,
    3.8715350915692364,
    3.7045949061144983,
]



x_train = x[1:8]
y_train = y[1:8]
x_test = x[9:end]
y_test = y[9:end]

f = GP(Matern52Kernel())

fx = f(x_train, 0.1)
logpdf(fx, y_train)


p_fx = posterior(fx, y_train)
logpdf(p_fx(x_test), y_test)


function gp_loglikelihood(x, y)
    function loglikelihood(params)
        kernel =
            softplus(params[1]) * (Matern52Kernel() ∘ ScaleTransform(softplus(params[2])))
        f = GP(kernel)
        fx = f(x, 0.1)
        return logpdf(fx, y)
    end
    return loglikelihood
end

const loglik_train = gp_loglikelihood(x_train, y_train)

## DHMC

using DynamicHMC
using LogDensityProblems

n_samples = 2_000
n_adapts = 1_000


# Log joint density
function LogDensityProblems.logdensity(ℓ::typeof(loglik_train), params)
    return ℓ(params) + logprior(params)
end
logprior(params) = logpdf(MvNormal(2, 1), params)

# The parameter space is two-dimensional
LogDensityProblems.dimension(::typeof(loglik_train)) = 2

# `loglik_train` does not allow to evaluate derivatives of
[...]

PR #334, opened 2022-09-17
Update the intro example

devmotion:
This PR updates the intro example.
In #333 I noticed we use a deprecated syntax for `MvNormal` and an internal field that was removed in DynamicHMC 3.3.
I think both things should be fixed.

Fixes #331. Fixes #330.


devmotion:
@willtebbutt Did you enable buildkite for the whole JuliaGP organization? That seems to cause CI errors, I think it should only be enabled for repos with a buildkite configuration.

devmotion:
PPL test failures are unrelated and caused by Soss (possibly fixed by updating it but IMO that should be done in a separate PR).

theogf (approved):

willtebbutt:
> @willtebbutt Did you enable buildkite for the whole JuliaGP organization? That seems to cause CI errors, I think it should only be enabled for repos with a buildkite configuration.

I enabled it for KernelFunctions, AbstractGPs, and ApproximateGPs. I need to make PRs to configure buildkite to basically do nothing on each of those repos. Apologies for the inconvenience, will add to my todo list!

PR #336, opened 2022-09-26
Improve docs

devmotion:
Improve the docs and fix build failures.

The initial commit probably does not fix the errors, I assume this requires to specify some slightly larger than default (1e-9) artificial observation noise for plotting (to fix numerical issues).

### DIFF
--- a/examples/3-parametric-heteroscedastic/script.jl
+++ b/examples/3-parametric-heteroscedastic/script.jl
@@ -79,7 +79,7 @@ with_theme(
         label="posterior + noise",
         color=(:orange, 0.3),
     )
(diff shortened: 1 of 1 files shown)


devmotion:
Seems the docs were built successfully even without adjusting the artificial noise. I assume the problem might reappear but maybe it's better to only adjust the noise level when the problem actually reappears?

devmotion:
The plot seems fine: https://juliagaussianprocesses.github.io/AbstractGPs.jl/previews/PR336/examples/3-parametric-heteroscedastic/

So this PR should be good to go (and we can increase the artificial noise in a follow-up PR if needed).

willtebbutt (approved):

PR #339, opened 2022-09-27
Do not re-evaluate the objective function

devmotion:
When working on #338, I noticed that potentially the optimization procedure in this example can be improved by not re-evaluating the objective function.

I assume if this turns out to be beneficial (I assume so but we'll see) then maybe the example in ParameterHandling could be updated as well or extended with a note in the tips and tricks section.

### DIFF
--- a/examples/3-parametric-heteroscedastic/script.jl
+++ b/examples/3-parametric-heteroscedastic/script.jl
@@ -76,7 +76,7 @@ result = optimize(
     ),
     Optim.Options(; show_every=100),
 )
-θ_final = unflatten(result.minimizer);
+θ_final = unflatten(result.minimizer)
 
 # Construct the posterior GP with the optimal model parameters:
 Σ_obs_final = observation_variance(θ_final, x)
@@ -93,6 +93,10 @@ with_theme(
         Axis=(limits=((0, 10), nothing),),
     ),
 ) do
+    # Fix numerical issues when computing the Cholesky decomposition of the covariance matrix
+    # of the finite projection of the posterior GP by introducing artifical noise
+    f_post_jitter = f_post(x, 1e-8)
+
     plot(
         x,
         f_post(x, Σ_obs_final);
@@ -100,8 +104,8 @@ with_theme(
         label="posterior + noise",
         color=(:orange, 0.3),
     )
-    plot!(x, f_post; bandscale=3, label="posterior")
-    gpsample!(x, f_post; samples=10, color=Set1_4[3])
+    plot!(x, f_post_jitter; bandscale=3, label="posterior")
+    gpsample!(x, f_post_jitter; samples=10, color=Set1_4[3])
     scatter!(x, y; label="y")
     axislegend()
     current_figure()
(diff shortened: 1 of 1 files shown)


devmotion:
It seems the potential performance gain is only minor. Maybe run time changed from ~18 to ~17 minutes but I'm not sure if that is really reproducible.

@theogf However, I think I found the underlying cause for the numerical issues that show up from time to time: Due to the scaling with `x^2` the observation variance in the model, as currently specified on the master branch, goes to zero as x -> 0, regardless of the value of the scaling hyperparameter. I added a lower bound which seemed to improve things.

devmotion:
I guess it's not too surprising that there is no significant performance gain. From https://juliagaussianprocesses.github.io/AbstractGPs.jl/previews/PR339/examples/3-parametric-heteroscedastic/:
```julia
 * Status: success

 * Candidate solution
    Final objective value:     1.576205e+02

 * Found with
    Algorithm:     L-BFGS

 * Convergence measures
    |x - x'|               = 3.97e-08 ≰ 0.0e+00
    |x - x'|/|x'|          = 1.58e-08 ≰ 0.0e+00
    |f(x) - f(x')|         = 0.00e+00 ≤ 0.0e+00
    |f(x) - f(x')|/|f(x')| = 0.00e+00 ≤ 0.0e+00
    |g(x)|                 = 6.34e-09 ≤ 1.0e-08

 * Work counters
    Seconds run:   1  (vs limit Inf)
    Iterations:    10
    f(x) calls:    13
    ∇f(x) calls:   11
```

devmotion:
I also noticed that we already use a similar approach with Optim.only_fg! in https://juliagaussianprocesses.github.io/AbstractGPs.jl/previews/PR339/examples/1-mauna-loa/

devmotion:
I think (some of) the changes in this PR are useful nevertheless (e.g., displaying success and timing of the optimization, a bit simpler code, non-zero variances), so I think the PR is ready for review.

examples/3-parametric-heteroscedastic/script.jl
```diff
@@ -21,51 +21,69 @@ using Random
 Random.seed!(42)  # setting the seed for reproducibility of this notebook
 #md nothing #hide
 
-# Specify simple GP:
-build_gp(θ) = GP(0, θ.s * with_lengthscale(SEKernel(), θ.l));
-
-# Observation variance is some scaling of $x^2$:
-observation_variance(θ, x::AbstractVector{<:Real}) = Diagonal(θ.σ² .* x .^ 2);
+# In this example we work with a simple GP with a Gaussian kernel and heteroscedastic observation variance.
+observation_variance(θ, x::AbstractVector{<:Real}) = Diagonal(0.01 .+ θ.σ² .* x .^ 2)
```
theogf, line 25: if `x^2` is an issue why not use `exp` ?

theogf (commented):
Looks much cleaner indeed. Maybe a small modelling improvement would be to use `exp` instead of `x^2` ?

devmotion:
I wanted to stick to the current definition of the observation variance as much as possible and only change it in a minimal way. In principle, of course, any function of `x` with values bounded away from zero should work. `exp` would not be my first choice though as it would require scaling and damping to achieve a similar behaviour as currently and in particular to avoid exploding variances for x >> 0. Maybe one could use `s * log1pexp(x)` to avoid the exponential growth.

theogf:
> I wanted to stick to the current definition of the observation variance as much as possible and only change it in a minimal way. In principle, of course, any function of `x` with values bounded away from zero should work. `exp` would not be my first choice though as it would require scaling and damping to achieve a similar behaviour as currently and in particular to avoid exploding variances for x >> 0. Maybe one could use `s * log1pexp(x)` to avoid the exponential growth.

Sure, I also have a version with `s * logistic(x)`, but that could be for another work.
LGTM !

PR #343, opened 2022-12-05
Fix broken link in docs

simsurace:
**Summary**
The link to the practical guide seems to be broken.

**Proposed changes**
I replaced the link by another one that works. I'm not sure whether this is in fact the same article, but from the author's list it seems plausible that this is in fact the same or a newer version.
Could someone who remembers the original confirm this?


willtebbutt (approved):
I don't entirely recall whether this was the original article, but it's definitely a good reference, so I'm happy to merge.

PR #349, opened 2022-12-28
Bump LogDensityProblems

willtebbutt:
**Summary**

Updates the `LogDensityProblems` and `AdvancedHMC` deps in our initial example. Supercedes #347 and #348 . There were some breaking changes and it seemed more straightforward to just handle all of the updates in a single go.

**Proposed changes**

Bump the deps, add `LogDensityProblemsAD` as a dep in the initial example.

**Breaking changes**

None


willtebbutt:
Seems to fix the problems associated with the breaking changes in `LogDensityProblems`

theogf (approved):

examples/0-intro-1d/script.jl
```diff
@@ -149,9 +150,13 @@ n_adapts = 1_000
 
 # Define a Hamiltonian system of the log joint probability.
 
-logjoint_train(params) = loglik_train(params) + logprior(params)
+struct LogJointTrain end
```
devmotion, line 153: Probably shouls unify this with the already existing LogDensityProblems interface impl for DynamicHMC below. With this PR, it seems now we have to different definitions of the log joint with LogDensityProblems. This seems a bit unnecessarily complicated.

willtebbutt: That's a good point -- I'm going to open an issue based on this comment, as I agree that it should probably be tidied up.

Issue #319, opened 2022-04-29
Should LatentFiniteGP contain the noise?


theogf:
Maybe I am missing the point but it feels like there is a discrepancy between `LatentGP` and `AbstractGP` although they kinda behave the same.
I don't understand why the noise should be built in `LatentGP` and not in `LatentFiniteGP` ?

Issue #318, opened 2022-04-19
VFE/DTC should be moved to ApproximateGPs.jl


theogf:

willtebbutt:
I keep changing my mind on this. On the one hand, the name suggests that this is true. On the other, VFE / DTC / FITC don't add any additional deps to AbstractGPs, and the approximate posteriors produced satisfy the AbstractGPs API.

I don't think I would get in the way of this move, but I can also live with the status quo.

theogf:
Well as a naive user comment, I kept looking in ApproximateGPs for it before realizing it was in AbstractGPs after asking on Slack.

willtebbutt:
Haha fair enough. If you didn't know it was there, we should probably just move it here.

ngiann:
Just to add my 2 cents (far more naive than @theogf): After quite some digging to figure out where does Stheno find VFE, I was quite surprised to find it in AbstractGPs.

Issue #241, opened 2021-11-08
log marginal likelihood objective and posterior GP


st--:
When preparing #240 I stumbled across having to pass both FiniteGP for input features `f(x)` and targets `y` to both `posterior` (for PosteriorGP) and `logpdf` (for log marginal likelihood). Maybe the answer is no, but should we have something like `logpdf(::PosteriorGP)`?

Also, is it sufficiently intuitive that you get the log marginal likelihood by calling logpdf? why would calling the probability density function return what we get when we integrate out the uncertainty over the GP?

willtebbutt:
> When preparing #240 I stumbled across having to pass both FiniteGP for input features f(x) and targets y to both posterior (for PosteriorGP) and logpdf (for log marginal likelihood). Maybe the answer is no, but should we have something like logpdf(::PosteriorGP)?

There's nothing to stop us doing this. It obviously can't be part of any of the AbstractGPs APIs, but that's not to say we shouldn't add a special case thing for `PosteriorGP`.

> Also, is it sufficiently intuitive that you get the log marginal likelihood by calling logpdf? why would calling the probability density function return what we get when we integrate out the uncertainty over the GP?

`logpdf(fx, y)` says "give me the log probability density of y under the fx, where fx is the marginal distribution over f at x`, which has always felt clear to me (i.e. the log marginal probability density is just a log probability density under the marginal distribution).

Do you have a particular concern?

willtebbutt:
I was thinking about this again in the context of approximate inference.

I think I would be in favour of adding a function called `log_evidence(::PosteriorGP)` and (something along the lines of ) `approx_log_evidence(::ApproxPosteriorGP)`.
This would give us the best of two worlds:
1. it would maintain compatibility with the Distributions interface, i.e. we'd retain `logpdf(fx, y)`, which is needed for compatibility with lots of things, and
2. we would resolve the issue of needing to recompute intermediate quantities if you want both the log evidence and access to the posterior.

It's also quite clear what `log_evidence` ought to mean in the context of a `PosteriorGP`.

What do you think @st-- ?

Issue #223, opened 2021-09-27
LatentGP API: should ApproxPosteriorGP also store the likelihood?


st--:
And/or should we have a separate LatentApproxPosteriorGP?...

willtebbutt:
To my mind `ApproxPosteriorGP` ought to store only what it needs in order to construct the GP corresponding to the approximate posterior.

I would be favour of `posterior(::LatentGP)` producing another `LatentGP`-like object (possibly a `LatentGP` containing an `ApproxPosteriorGP` or a new type if that doesn't work out for some reason).

st--:
You're right, we can just construct `LatentGP(ApproxPosteriorGP(...), ...)` - e.g. as follows (excerpted from https://github.com/JuliaGaussianProcesses/ApproximateGPs.jl/pull/59/files#diff-2581bfaaa21433d043e4343891c3547bbba68f31c35416bd6fcc9773e256c673R333):
```julia
function AbstractGPs.posterior(la::LaplaceApproximation, lfx::LatentFiniteGP, ys)
    dist_y_given_f, K, newton_kwargs = _check_laplace_inputs(lfx, ys; la.newton_kwargs...)
    _, cache = _newton_inner_loop(dist_y_given_f, ys, K; newton_kwargs...)
    f_post = ApproxPosteriorGP(la, lfx.fx, cache)
    return LatentGP(f_post, lfx.lik, lfx.fx.Σy)
end
```
The one really annoying thing is that `LaplaceGP` requires an explicit jitter argument, and it seems weird to have to access `lfx.fx.Σy` - which should be an implementation detail this piece of code shouldn't need to know about... what do you think?

willtebbutt:
> The one really annoying thing is that LaplaceGP requires an explicit jitter argument, and it seems weird to have to access lfx.fx.Σy - which should be an implementation detail this piece of code shouldn't need to know about... what do you think?

That's a good point -- it is annoying. IIRC this is just a limitation of our abstractions at the minute. I _think_ we can be fairly confident that `lfx.fx.Σy` does indeed represent a jitter term rather than anything more complicated, so hopefully it's okay...

willtebbutt:
[...]

PR #361, opened 2023-04-25
Clarify API for GP approximations

### LINKED ISSUE #221: Add API for (log) marginal likelihood approximation
AbstractGPs currently defines `elbo`, but other approximations (e.g. Laplace, EP) also provide approximations to the (log) marginal likelihood that can be used for hyperparameter optimisation. Would be great to have a unified API function for that (for VFE, SVGP, etc it could simply forward to `elbo`). What do you think? Finding a decent name seems the most complicated aspect of this:
- `approx_lml` (short but a bit cryptic)?
- `approx_log_marginal_likelihood` (explicit but so long)?
- `approx_log_marglik` (the compromise that makes everyone unhappy)?
- `approx_log_evidence`?

st--:
**Summary**
AbstractGPs.jl should define the API we expect across the ecosystem. Currently, this is not the case for GP approximations, even though they're already introduced in this package (`VFE` for sparse GPR).

**Proposed changes**
Expose `posterior` and `approx_log_evidence` as part of the API with nice generic docstrings explaining their purpose.

**Related issues**
Resolves #221 

Also see #223, #241, #318, maybe #319

Status quo: some definitions in ApproximateGPs.jl, but people not realizing that you could just use a three-parameter form for `posterior`, leading to more reimplementation of base FiniteGPs than would be necessary (e.g. https://github.com/SebastianCallh/IterGP.jl/blob/AbstractGPs-interface/src/gp.jl)


st--:
> Should there be an `AbstractGPApproximation` type or so? That other packages can/ should use?

Hm... where would that be useful? Is there anything that you would be able to do with _any_ approximation, without knowing what approximation it is (i.e. it'd be necessary as a common base type in the hierarchy)? Or just as a "type hint" (i.e. it might be better to just improve the docs)?

st--:
Open question from https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/issues/221#issuecomment-1521960655:

should we add a `struct ExactGP end` "approximation" type to support a consistent API regardless of whether one is using exact GPR or a sparse approximation such as VFE/DTC?

Crown421:
> Open question from [#221 (comment)](https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/issues/221#issuecomment-1521960655):
> 
> should we add a `struct ExactGP end` "approximation" type to support a consistent API regardless of whether one is using exact GPR or a sparse approximation such as VFE/DTC?

I think this makes sense to me, with appropriate aliases for the current two argument form. 

Re the above, I was thinking about a type hint, need to think about whether such a type might also be useful elsewhere.

Crown421:
One way I can think of where a `AbstractApproximation` type might be useful in starting a more general hierarchy. 
Next level down would be the methods mentioned here: https://github.com/JuliaGaussianProcesses/ApproximateGPs.jl/issues/84, might might be some supertype of `VFE`, `DTC` (possibly others in the future like FITC)

src/exact_gpr_posterior.jl
```diff
@@ -3,6 +3,12 @@ struct PosteriorGP{Tprior,Tdata} <: AbstractGP
     data::Tdata
 end
 
+struct ExactGP end
+
+posterior(::ExactGP, fx::FiniteGP, y::AbstractVector{<:Real}) = posterior(fx, y)
```
theogf, line 8: I would have gone the other way around?

But maybe that's too complicated...

Crown421: I find this order of arguments quite intuitive.

src/sparse_approximations.jl
```diff
@@ -228,14 +245,18 @@ true
...
-function elbo(vfe::VFE, fx::FiniteGP, y::AbstractVector{<:Real})
+function approx_log_evidence(vfe::VFE, fx::FiniteGP, y::AbstractVector{<:Real})
     @assert vfe.fz.f === fx.f
-    _dtc, A = _compute_intermediates(fx, y, vfe.fz)
-    return _dtc - (tr_Cf_invΣy(fx, fx.Σy) - sum(abs2, A)) / 2
+    dtc_objective, A = _compute_intermediates(fx, y, vfe.fz)
+    return dtc_objective - (tr_Cf_invΣy(fx, fx.Σy) - sum(abs2, A)) / 2
 end
 
+elbo(vfe::VFE, fx, y) = approx_log_evidence(vfe, fx, y)  # TODO deprecate?
```
st--, line 254: Should we deprecate `elbo` or leave it as is (for `VFE`)?

simsurace: I like the more specific name as a synonym wherever it is applicable.

src/exact_gpr_posterior.jl
st--: Does it make any difference in practice? (I kinda like it as it is but that might just be status-quo bias too...)

Crown421: In practice I think the answer is no. 
Within the code, there might be something to be said for consistency. Every `posterior` is defined with the 3-argument form, but the exact one gets a special alias.

src/sparse_approximations.jl
```diff
@@ -13,7 +13,14 @@ struct VFE{Tfz<:FiniteGP}
     fz::Tfz
 end
 
-const DTC = VFE
+"""
+    DTC(fz::FiniteGP)
+
+Similar to `VFE`, but uses a different objective for `approx_log_evidence`.
```
st--, line 19: Could maybe do with a better docstring but then it needs to be sorted out more thoroughly anyways (see #309) and I can't  think of what it should be right now, so would leave that for some other PR/person/time...

test/sparse_approximations.jl
```diff
@@ -1,133 +1,136 @@
...
+            # Ensure that the elbo/dtc objective is close to the logpdf when appropriate.
+            @test approx_log_evidence(ApproxType(f(x)), fx, y) isa Real
+            @test approx_log_evidence(ApproxType(f(x)), fx, y) ≈ logpdf(fx, y)
+
+            if ApproxType === VFE
+                @test elbo(ApproxType(f(x)), fx, y) ==
+                    approx_log_evidence(ApproxType(f(x)), fx, y)
+                @test elbo(ApproxType(f(x .+ randn(rng, N_cond))), fx, y) < logpdf(fx, y)
+            end
+        end
```
st--, line 101: this is the main change

src/exact_gpr_posterior.jl
theogf: To be fair, there is only one posterior. Shouldn't VFE and DTC dispatched on approx_posterior?

st--: @theogf what do you mean? there is no `approx_posterior` method..

st--:
Hello! What do you think needs doing/addressing/... before merging this ?

src/sparse_approximations.jl
```diff
@@ -161,48 +170,56 @@ function update_posterior(f_post_approx::ApproxPosteriorGP{<:VFE}, fz::FiniteGP)
...
         Σy=f_post_approx.data.Σy,
     )
-    return ApproxPosteriorGP(VFE(fz_new), f_post_approx.prior, cache)
+    return ApproxPosteriorGP(
+        _update_approx(f_post_approx.approx, fz_new), f_post_approx.prior, cache
+    )
 end
 
+_update_approx(vfe::VFE, fz_new::FiniteGP) = VFE(fz_new)
+_update_approx(dtc::DTC, fz_new::FiniteGP) = DTC(fz_new)
```
st--, line 179: is this the right way to handle this? if anyone can think of a better approach please say:)

src/exact_gpr_posterior.jl
theogf: haha somehow in my mind there was a `approx_posterior` method. So yeah then it's back to

- Should we have a unique 3-args posterior API (where the 2-args default to `ExactInference`) ?
- Should we just have 2-args methods and have the GP wrapped (like in ApproximateGPs) ?

st--: ApproximateGPs has the 3-args posterior though, no wrapping?

src/exact_gpr_posterior.jl
```diff
@@ -3,10 +3,16 @@ struct PosteriorGP{Tprior,Tdata} <: AbstractGP
     data::Tdata
 end
 
+struct ExactGPR end
```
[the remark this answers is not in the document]
willtebbutt: I'd prefer `ExactInference`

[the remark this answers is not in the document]
st--: done

willtebbutt (approved):

Issue #378, opened 2023-09-19
Log-likelihood is lower after fitting the process?


filipporemonato:
I obtain a much lower (several orders of magnitude) log-likelihood after fitting the process to the data, even though the process seems to do a fit of the data correctly from the plots.
I expected the log-likelihood of the train data to increase after the GP is fitted to them.

This is exactly the same behaviour as in your doc's example: https://juliagaussianprocesses.github.io/AbstractGPs.jl/stable/examples/0-intro-1d/

Note that your log-likelihood is -25 for the un-fitted GP, and then becomes -232 after fitting.

Since the log-likelihood is, by its name, the log of a probability, and probabilities are in [0,1], the log-likelihood after fitting should be closer to 0 = log(1), instead of becoming even more negative.

willtebbutt:
Thanks for opening this. This is partly (kind of) a typo in the script, and partly that these two quantities are slightly different things. To replicate:
```julia
using AbstractGPs, Random

x = [
    0.8658165855998895,
    0.6661700880180962,
    0.8049218148148531,
    0.7714303440386239,
    0.14790478354654835,
    0.8666105548197428,
    0.007044577166530286,
    0.026331737288148638,
    0.17188596617099916,
    0.8897812990554013,
    0.24323574561119998,
    0.028590102134105955,
]
y = [
    1.5255314337144372,
    3.6434202968230003,
    3.010885733911661,
    3.774442382979625,
    3.3687639483798324,
    1.5506452040608503,
    3.790447985799683,
    3.8689707574953,
    3.4933565751758713,
    1.4284538820635841,
    3.8715350915692364,
    3.7045949061144983,
]

x_train = x[1:8]
y_train = y[1:8]
x_test = x[9:end]
y_test = y[9:end]

f = GP(Matern52Kernel())

fx = f(x_train, 0.1)
logpdf(fx, y_train)

p_fx = posterior(fx, y_train)
logpdf(p_fx(x_test), y_test)
```
Firstly, the last statement here should really have the noise term in it:
```julia
logpdf(p_fx(x_test, 0.1), y_test)
```

Secondly, the log joint probabilities of all the training data and all of the test data aren't comparable quantities, because there's different amounts of data in each data set, and they're quite different subsets of the data. We're better off comparing the logpdf of the train set under the prior / posterior, and test set under the prior / posterior:
```julia
julia> logpdf(fx, y_train)
-25.53057444906228

[...]

PR #379, opened 2023-09-19
Fix typo in intro example

willtebbutt:
**Summary**

Fixes noise-related problem raised in #378 

**Proposed changes**

Add the correct noise variance wherever posterior log marginal likelihoods are asked for.

None

**Breaking changes**

N/A


st-- (approved):
lgtm... I'm just thinking if it'd be more helpful to have an explicit variable `noise_var = 0.1` and then use that in the arguments ?

willtebbutt:
> lgtm... I'm just thinking if it'd be more helpful to have an explicit variable noise_var = 0.1 and then use that in the arguments ?

This is a good point. I'll modify this, then merge when CI passes.

PR #390, opened 2024-02-10
Enable periodic kernel in example

simsurace:
Now that https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/531 has been merged, we can try re-enabling the `PeriodicKernel`.


willtebbutt (approved):
This looks to me to be working from the docs that have been built during CI. @simsurace do you agree?

simsurace:
Yeah everything works, I tested locally as well. There is also a unit test for AD through the periodic kernrl.

PR #391, opened 2024-02-12
Create Downgrade-CI and bump lower compat bounds

simsurace:
Following the general suggestion [here](https://discourse.julialang.org/t/psa-add-downgrade-ci-to-better-check-version-compatability/110063). I'm trying the workflow file from [here](https://gist.github.com/aplavin/ec1024d00db6ebe8a7fa0a38aa8cca6c) to speed up finding bad lower bounds.

The PR raises lower bounds such that if any single one of the dependencies is pinned at the lower bounds
- The remaining dependencies can still be resolved
- All the tests pass

While this does not guarantee passing tests if multiple dependencies are held back at once, in my view it is an improvement over the status quo.


simsurace:
I guess this means some of the lower bounds are not outdated/no longer working.

simsurace:
Ok, these are the highest lower bounds that work.

test/Project.toml
```diff
@@ -11,18 +11,3 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
...
-FiniteDifferences = "0.9.6, 0.10, 0.11, 0.12"
-LinearAlgebra = "1"
-PDMats = "0.11"
-Pkg = "1"
-Plots = "1"
-Random = "1"
-Statistics = "1"
-Zygote = "0.5, 0.6"
-Test = "1"
-julia = "1.6"
```
theogf, line 28: Any reason to remove all test compats?

theogf: Or is it still for the exploration phase?

simsurace: Still exploring :)

simsurace:
I think I went about it the wrong way. I bumped the lower bounds before the pipeline was doing the correct thing. I should maybe start from scratch.

test/Project.toml
simsurace: The thing is, the action only downgrades compats in the main Project.toml. So the ones in the tests interfere. Once the main ones are set, we can copy them over to the test ones (if we want).

simsurace:
Ok I think now we have it. Switching to the low-cost action now.

simsurace:
I think the action that pins all dependencies at their minima is too restrictive. I will disable both by default and let them be enabled by the `downgrade-ci` PR flag. Let us then decide on a case-by-case basis.

theogf (commented):
I don't think I fully understand the action? Is the idea to see which is the lowest compatible version of our dependencies?

theogf:
> I don't think I fully understand the action? Is the idea to see which is the lowest compatible version of our dependencies? 

I am asking this because it looks like the result is much more restrictive than before

simsurace:
The action fixes each dependency separately to its lowest version that still resolves, and runs tests there. In response, I bumped the lower bound to the lowest possible version where the test passes. Sure, it is more restrictive, but the looser bounds from before are broken.

simsurace:
In contrast, the test-simple job would pin all dependencies to their lower bounds simultaneously and then runs tests there. I figured that this is too strict a requirement.

yebai:
Please feel free to open another PR if this is still needed.

PR #396, opened 2024-03-24
fix a docstring typo in intro

lanceXwq:


willtebbutt (approved):
Good catch -- thanks!

PR #403, opened 2024-07-04
Gaussion Processes : added navbar workflow

jeetsuthar:
We are adding this to all Turing satellite packages! See here - https://github.com/TuringLang/turinglang.github.io/blob/main/assets/scripts/README.md

We also announced this navigation bar in our Turing slack's general channel!

### DIFF
--- a/.github/workflows/DocsNav.yml
+++ b/.github/workflows/DocsNav.yml
@@ -1,0 +1,44 @@
+name: Add Navbar
+
+on:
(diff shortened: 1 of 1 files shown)


willtebbutt:
Sorry, it's not clear to me what the purpose of this is. Can I see an example of what this looks like in the existing docs somewhere before we commit to it in the JuliaGPs ecosystem (which is separate from the Turing.jl ecosystem)

PR #415, opened 2025-02-20
Resolve issues with Zygote.jl and `mean_vector`

soldasim:
**Summary**

There is an issue with Zygote.jl not being able to correctly differentiate through `mean_vector`. It is described in [the docstring of `CustomMean`](https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/blob/8d431a2ed31e3bf937640948c87aa85948ab8688/src/mean_function.jl#L42C1-L66C60). It proposes to the user to implement custom methods for their particular mean function if they wish to use Zygote.jl. This seems unnecessarily clumsy.

**Proposed changes**

Correct me if I'm wrong, but I think this issue can be resolved by simply defining special methods for `ColVecs` and `RowVecs` as proposed in this PR. This way, Zygote.jl should work with `CustomMean` out of the box.

nothing

**Breaking changes**

This should not introduce any breaking changes to users who already resolved the issue by defining their own methods as described in the current docstring.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "AbstractGPs"
 uuid = "99985d1d-32ba-4be9-9821-2ec096f28918"
 authors = ["JuliaGaussianProcesses Team"]
-version = "0.5.21"
+version = "0.5.22"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/src/mean_function.jl
+++ b/src/mean_function.jl
@@ -44,23 +44,12 @@ mean_vector(m::ConstMean, x::AbstractVector) = Fill(m.c, length(x))
 
 A wrapper around whatever unary function you fancy. Must be able to be mapped over an
 `AbstractVector` of inputs.
-
-# Warning
-`CustomMean` is generally sufficient for testing purposes, but care should be taken if
-attempting to differentiate through `mean_vector` with a `CustomMean` when using
-`Zygote.jl`. In particular, `mean_vector(m::CustomMean, x)` is implemented as `map(m.f, x)`,
-which when `x` is a `ColVecs` or `RowVecs` will not differentiate correctly.
-
-In such cases, you should implement `mean_vector` directly for your custom mean.
-For example, if `f(x) = sum(x)`, you might implement `mean_vector` as
-```julia
-mean_vector(::CustomMean{typeof(f)}, x::ColVecs) = vec(sum(x.X; dims=1))
-mean_vector(::CustomMean{typeof(f)}, x::RowVecs) = vec(sum(x.X; dims=2))
-```
-which avoids ever applying `map` to a `ColVecs` or `RowVecs`.
 """
 struct CustomMean{Tf} <: MeanFunction
     f::Tf
 end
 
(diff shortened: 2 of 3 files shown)


willtebbutt:
Thanks for this @soldasim . Could you add a couple of tests to check that it's working correctly?

I believe that the original reason that we introduced this was precisely because Zygote struggles with the kind of implementation that you propose (really bad performance), so we opted to not provide implementations in general. As such, it would also be good to see a benchmark that shows that Zygote's performance isn't completely terrible for this implementation.

soldasim:
Hi @willtebbutt,

I've added some a unit test in `test/mean_function.jl` for the two new methods.

I've also added [a problem](https://discourse.julialang.org/t/zygote-gradient-does-not-work-with-abstractgps-custommean/87815) that I've had in the past with Zygote as a test. This ensures that the new methods really fix the issue. I cannot be sure that different issues with Zygote.jl gradients could not exist, which may not be fixed by the two new methods though.

soldasim:
I have tested the performance on the following benchmark;

<details>
  <summary>benchmark</summary>
  
```julia
using AbstractGPs
using Random
using BenchmarkTools
using Zygote

rng = MersenneTwister(123456)
N, D = 5, 3
xD_colvecs = ColVecs(randn(rng, D, N))
xD_rowvecs = RowVecs(randn(rng, N, D))

X = randn(rng, D, N)
y = randn(rng, N)

foo_mean = x -> sum(abs2, x)

function construct_finite_gp(X, lengthscale, noise)
    mean = CustomMean(foo_mean)
    kernel = with_lengthscale(Matern52Kernel(), lengthscale)
    return GP(mean, kernel)(X, noise)
end

function loglike(lengthscale, noise)
    gp = construct_finite_gp(X, lengthscale, noise)
    return logpdf(gp, y)
end

# # User-defined methods.
# # Comment/Uncomment to compare the proposed implementation with user-defined methods.
# # (And restart REPL.)
# AbstractGPs.mean_vector(m::CustomMean{typeof(foo_mean)}, x::ColVecs) = vec(sum(abs2, x.X; dims=1))
# AbstractGPs.mean_vector(m::CustomMean{typeof(foo_mean)}, x::RowVecs) = vec(sum(abs2, x.X; dims=2))

display(@btime Zygote.gradient(n -> loglike(1., n), 1.))
display(@btime Zygote.gradient(l -> loglike(l, 1.), 1.))
```

</details>

Results with the proposed implementation:

```julia-repl
julia> 

  56.417 μs (697 allocations: 73.02 KiB)
(6.037955692260997,)
  56.709 μs (697 allocations: 73.02 KiB)
(5.834500256839596,)

julia> 
```

Results with the user-defined functions:

```julia-repl
julia> 

  55.625 μs (679 allocations: 69.52 KiB)
(6.037955692260998,)
  55.917 μs (679 allocations: 69.52 KiB)
(5.834500256839596,)

julia> 
```

At least in this case, the proposed implementation is as efficient as defining custom functions according to the old docstring.

willtebbutt (commented):
This looks to be nearly there now. Could you please bump the patch version, and also move the tests that you've added to `zygote.jl` in to `test/mean_function.jl` instead?

willtebbutt (approved):
Thanks @soldasim . Docs not building does not appear to have anything to do with this PR, so I'm going to merge + tag.

PR #421, opened 2025-04-06
GP Sampling

Crown421:
**Summary**
This PR is a (at this stage) draft PR to allow sampling of GP, including sampling via Cholesky, via Random Fourier Features, and pathwise/ decoupled sampling. 

**Proposed changes**
The main interface revolves around a new `GPSampler` struct, as a subtype of `Random.Sampler, which takes a GP and a "method". One can then draw from this sampler using `rand`. 

However, there are still issues to this implementation
- GP must have a kernel (when using RFF)
- Not yet working with sparse approximations
- Pathwise functionality was written to (theoretically) enable using Cholesky samples for the prior, and RFF for the posterior, but in practice one would never want to do that. 
   - It might not be worth to keep this use case around
- Multi-Output probably not working yet (and not tested)

Making a separate package, and/or (due to the complexity) just keeping it good enough for my needs.

**Breaking changes**
Should not be breaking (at this point).


willtebbutt:
Very happy to give this a review -- when you're ready for one, please do tag me.

Crown421:
@willtebbutt If you have some time, I would already appreciate a rough review, on the overall structure/ idea. 
I worry that I got too "clever" with the type structure/ relationships/ interface.

Project.toml
```diff
@@ -9,6 +9,7 @@ Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
 FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
 IrrationalConstants = "92d709cd-6900-40b7-9082-c6be49f344b6"
 KernelFunctions = "ec8451be-7e33-11e9-00cf-bbf324bd1392"
+KernelSpectralDensities = "027d52a2-76e5-4228-9bfe-bc7e0f5a8348"
```
yebai, line 12: Can you move this into a package extension?

Crown421: Apologies, it has been a minute. 

The short answer is that I am not sure. On the one hand, it would make sense to keep everything clean, but
- KernelSpectralDensities is a tiny package
- The main reason it exists is to enable decoupled sampling with Fourier basis functions, which (AFAIK) still the state of the art for sampling GPs
- (the main uncertainty for me) I am not sure where to put some of the objects, since they only make sense when using the combination of both packages.

